PR #30, opened 2023-01-02
17 connect existing nodes

### LINKED ISSUE #17: Connect existing nodes
Frequently, two problems can be partially solved (to different extents) by the same two solutions. Without being able to connect existing nodes, there's no way to model this:
image

I want to be able to click and drag between two node handles to create an edge between them. This should only be enabled in the problem mapping - since claims are in a tree, they always have one parent and therefore don't need this functionality.

**Additional context**

**Technical ideas**
See [react-flow example](https://reactflow.dev/docs/getting-started/adding-interactivity/#handle-connections) for how do this (note: use the diagram store instead of the `setNodes` hooks that are used in the example).

Will also probably need a new [diagram store action](https://github.com/keyserj/ameliorate/blob/main/web/src/modules/topic/store/actions.ts) like `connectNodes(node1, node2)`.

keyserj:
Closes #17 

### Description of changes

- see commit messages

### Additional context

-

### DIFF
--- a/web/src/modules/topic/components/AddNodeButton/AddNodeButton.tsx
+++ b/web/src/modules/topic/components/AddNodeButton/AddNodeButton.tsx
@@ -1,6 +1,7 @@
 import { addNode } from "../../store/actions";
 import { type RelationDirection } from "../../utils/diagram";
-import { NodeType, RelationName, nodeDecorations } from "../../utils/nodes";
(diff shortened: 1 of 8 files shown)


keyserj:
Label collision avoidance isn't great but that's not a huge deal right now. Created #31 to address

PR #49, opened 2023-02-23
Added titles and aria-labels to menu icons without text

Nezzely:
### Description of changes

- For the following Menu buttons: [undo, redo, download, upload, reset]. Added text descriptions when hovered  and audio readout.

### Additional context

-  Gives additional context to button actions for hearing impaired & user hover actions.

### DIFF
--- a/web/src/modules/topic/components/TopicToolbar/TopicToolbar.tsx
+++ b/web/src/modules/topic/components/TopicToolbar/TopicToolbar.tsx
@@ -59,10 +59,10 @@ export const TopicToolbar = () => {
           <MenuItem onClick={() => loadExample("world_hunger.json")}>World Hunger</MenuItem>
           <MenuItem onClick={() => loadExample("ameliorate.json")}>Ameliorate</MenuItem>
         </Menu>
-        <IconButton color="inherit" onClick={downloadTopic}>
+        <IconButton color="inherit" title="Download" aria-label="Download" onClick={downloadTopic}>
           <Download />
         </IconButton>
-        <IconButton color="inherit" component="label">
+        <IconButton color="inherit" component="label" title="Upload" aria-label="Upload">
           <Upload />
           <input hidden accept=".json" type="file" onChange={uploadTopic} />
         </IconButton>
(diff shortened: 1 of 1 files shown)


web/src/modules/topic/components/TopicToolbar/TopicToolbar.tsx
```diff
@@ -59,10 +59,10 @@ export const TopicToolbar = () => {
           <MenuItem onClick={() => loadExample("world_hunger.json")}>World Hunger</MenuItem>
           <MenuItem onClick={() => loadExample("ameliorate.json")}>Ameliorate</MenuItem>
         </Menu>
-        <IconButton color="inherit" onClick={downloadTopic}>
+        <IconButton color="inherit" title="Download" aria-label="Download" onClick={downloadTopic}>
```
keyserj, line 62: nice usage of `title` - I was thinking a tooltip component would have to be used but this seems to do the trick nicely.

Was trying to figure out if `aria-label` is necessary if `title` is set... TIL about devtools' [accessibility tab](https://learn.microsoft.com/en-us/microsoft-edge/devtools-guide-chromium/accessibility/accessibility-tab), which actually shows that the Name property is being set without aria-label. However, a few different sources ([random blog post](https://inclusive-components.design/tooltips-toggletips/), [stack overflow](https://stackoverflow.com/a/63726152)) critique sole usage of `title`, so I think it's good to set `aria-label` as well to be safe.

keyserj (approved):
One note, not a big deal - I've been trying to use [conventional commits](https://github.com/amelioro/ameliorate/blob/main/CONTRIBUTING.md#conventional-commits) to improve readability of the commit history (there's also tooling that can take advantage of it, though I haven't set any up yet...). A commit message for this in that format would be like "touchup: add titles and aria-labels to menu icons".

Nice work

seal

keyserj:
Ayy this increased the site's lighthouse accessibility score from 84 to 87!
before https://app.netlify.com/sites/velvety-vacherin-4193fb/deploys/63f54ed019ca060008f01a35
after https://app.netlify.com/sites/velvety-vacherin-4193fb/deploys/63f7e0581ea1730008d3b09e

Issue #19, opened 2022-12-30
Improve score UX


keyserj:
**Is your feature request related to a problem? Please describe.**
Right now, the score UX has a few issues:

1. no visual indication of the score besides the text itself - a color and/or more meaningful shape indication would be good. this is important because a big mapping should be able to draw attention based on score values.
2. have to click a small, awkwardly-floating button to select a score
3. score button has unintuitively differing hover & click functionality

Some important other requirements:

4. without being interacted with, the score UX needs to fit in the corner of a node and on the center of an edge
5. when interacted with, the score UX _can_ grow to fit more information and/or make it easier to further interact with. It likely would need to lay in front of anything around it.
6. edges will eventually also have some text on them (e.g. "solves" "causes"), so score UX needs to fit with this
7. should be able to remove score selection (right now there's a "-") to distinguish a low/average score from no score

**Describe the solution you'd like**
A separate button for viewing the child claim tree could address number 3 (and could use the outlined, single, multi versions of the icon to indicate how many children claims are in the tree too):
![image](https://user-images.githubusercontent.com/13872370/210022635-d0d0191e-537a-4908-b452-deed844f69c7.png)

For addressing number 1, perhaps each number could change the button's color across some gradient. Color (same color?) could also be used with a shadow (scaling thickness/boldness?) around the relevant node/edge.

For addressing number 2, perhaps the button should expand into a [slider](https://mui.com/material-ui/react-slider/) (though I'm not sure how this would solve number 7)? Or a [pie menu](https://github.com/psychobolt/react-pie-menu) (ideally though we wouldn't use a component that's unpopular)?

**Describe alternatives you've considered**

**Additional context**

**Technical ideas**

keyserj:
Currently thinking that this with a pie menu for choice seems pretty good
![image](https://user-images.githubusercontent.com/13872370/218867982-80f752f4-35f2-4116-9ba2-6e7212584ec2.png)

PR #75, opened 2023-03-06
Improve score ux

keyserj:
Partially addresses #19 . Still would like to:
- use pie menu
- add color to scores based on the number
- reduce score options to 1-9
- maybe try a square/icon button instead of custom circular button

### Description of changes

- extract claim indicator for viewing claim diagram (score on-click now just opens score options)
- add box to edge UX
- combined many components into fewer directories (Node/, Indicator/, CriteriaTable/, Surface/)
- rest of commits are small refactors/touchups

### Additional context

-


web/src/modules/topic/components/CriteriaTable/CriteriaTable.styles.tsx
```diff
@@ -2,8 +2,8 @@ import { css } from "@emotion/react";
 import styled from "@emotion/styled";
 import { Button, IconButton } from "@mui/material";
 
-import { AddNodeButton } from "../AddNodeButton/AddNodeButton";
-import { nodeWidth } from "../EditableNode/EditableNode.styles";
+import { AddNodeButton } from "../Node/AddNodeButton";
```
keyserj, line 5: FYI @Nezzely this last commit tries to reorganize the `modules/topic/components/` directory since it was getting pretty big. I don't think it's perfect but it seems maybe better for now? Lemme know if you have any suggestions

PR #85, opened 2023-03-13
fix: Score Dial appropriately closes

### LINKED ISSUE #81: Styled Popper does not disappear on the corners.
**Describe the bug**
Expected ScorePie component  to disappear when a score is selected or when a user moves their mouse outside of the circular dial.
When a use moves their mouse in a X pattern centered on the ScorePie the dial doesnt close when the user exits for the "corners".

**To Reproduce**
1. Go to a node
2. Click on Scoring number
3. Move mouse to 1 of the 4 corners of the bounding viewing box
4. Expected the ScorePie to disappear.

**Expected behavior**
- When a mouse is leaving the circle of the ScorePie, I expect it to disappear when mouse leaves the edge from every angle.

**Screenshots**
Expected: 
- bounding box for the mouse leave event.
- image

Actual:
- bounding box for the mouse leave event
- image

**Additional context**
- The component uses [react-minimal-pie-chart](https://www.npmjs.com/package/react-minimal-pie-chart)
- The ScorePie has a viewing-box that is larger then the pie menu. 
- The mouse leave events fire for the whole bounding box of the component for the react-minimal-pie-chart.
   - Assumption is that the inner pie shards extend to the viewing box for the mouse leave event handler.
- The mouse leave event is actually a square.

**Technical ideas**
- Change the <Box> component to a CIRCLE that matches the inside radius circle of the react-minimal-pie-chart,
- Center the react-minimal-pie-chart within the <CircleBox> and move mouse leave events to it.

Nezzely:
Closes #81

### Description of changes

- Added functionality to segment mouse events,handles mouseleave and mouse hover events.
- Applies CSS to currently hovered segment.
- Selection wheel closes as soon as mouse leaves the pie chart.

### DIFF
--- a/web/src/modules/topic/components/Score/ScorePie.tsx
+++ b/web/src/modules/topic/components/Score/ScorePie.tsx
@@ -1,5 +1,5 @@
 import { useTheme } from "@emotion/react";
-import { Box, type PaletteColor } from "@mui/material";
+import { type PaletteColor } from "@mui/material";
 import { useState } from "react";
 import { PieChart } from "react-minimal-pie-chart";
 import { Data } from "react-minimal-pie-chart/types/commonTypes";
@@ -7,6 +7,7 @@ import { Data } from "react-minimal-pie-chart/types/commonTypes";
 import { setScore } from "../../store/actions";
 import { ArguableType, Score, possibleScores } from "../../utils/diagram";
 import { scoreColors } from "./Score";
+import { StyledBox } from "./Score.styles";
 
 interface Props {
   pieDiameter: number;
@@ -18,40 +19,42 @@ export const ScorePie = ({ pieDiameter, arguableId, arguableType }: Props) => {
   const theme = useTheme();
   const [hovered, setHovered] = useState<number | undefined>(undefined);
 
-  const data: Data = possibleScores.map((score, index) => {
+  const data: Data = possibleScores.map((score, dataIndex) => {
     const scoreColor = scoreColors[score];
     const paletteColor = theme.palette[scoreColor] as PaletteColor; // not sure how to guarantee this type, since palette has non-PaletteColor keys
 
     return {
       value: 1,
-      color: hovered === index ? paletteColor.dark : paletteColor.main,
+      color: hovered === dataIndex ? paletteColor.dark : paletteColor.main,
       key: score,
     };
   });
 
   return (
-    <Box display="flex" width={pieDiameter} height={pieDiameter}>
+    <StyledBox display="flex" width={pieDiameter} height={pieDiameter}>
       <PieChart
         data={data}
-        radius={50} // is relative to viewbox size [100, 100]
+        radius={45} // is relative to viewbox size [100, 100]
         label={({ dataEntry }) => dataEntry.key}
         labelStyle={() => ({
           fontSize: "14", // is relative to viewbox size [100, 100]
           pointerEvents: "none",
+          userSelect: "none",
         })}
-        segmentsStyle={{
+        segmentsStyle={(dataIndex) => ({
           transition: theme.transitions.create("all", {
             duration: theme.transitions.duration.shortest,
           }),
+          pointerEvents: "auto",
           cursor: "pointer",
-        }}
+        })}
         labelPosition={70} // percent of radius
         startAngle={-90 - 360 / possibleScores.length / 2} // shift first slice to top center
-        onMouseOver={(_, index) => setHovered(index)}
-        onMouseOut={() => setHovered(undefined)}
+        onMouseOver={(_, dataIndex) => setHovered(dataIndex)}
+        onMouseOut={(_) => setHovered(undefined)}
         onClick={(_, dataIndex) => setScore(arguableId, arguableType, data[dataIndex].key as Score)}
         background="white"
       />
-    </Box>
+    </StyledBox>
   );
 };
--- a/web/src/modules/topic/components/Score/ScorePie.tsx
+++ b/web/src/modules/topic/components/Score/ScorePie.tsx
@@ -42,6 +42,8 @@ export const ScorePie = ({ pieDiameter, arguableId, arguableType }: Props) => {
           userSelect: "none",
         })}
         segmentsStyle={(dataIndex) => ({
+          transform: hovered === dataIndex ? "scale(1.1)" : "",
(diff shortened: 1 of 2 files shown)


web/src/modules/topic/components/Score/ScorePie.tsx
```diff
@@ -30,28 +43,33 @@ export const ScorePie = ({ pieDiameter, arguableId, arguableType }: Props) => {
...
       <PieChart
         data={data}
-        radius={50} // is relative to viewbox size [100, 100]
+        radius={45} // is relative to viewbox size [100, 100]
         label={({ dataEntry }) => dataEntry.key}
         labelStyle={() => ({
           fontSize: "14", // is relative to viewbox size [100, 100]
           pointerEvents: "none",
+          webKitUserSelect: "none" /* Safari */,
+          msUserSelect: "none" /* IE 10 and IE 11 */,
```
keyserj, line 55: yooooo idk, mozilla doesn't even show IE in their browser compatibility list anymore, which makes me think it's not worth supporting (we'd have to go elsewhere for compatibility docs i guess?). what do you think?

web/src/modules/topic/components/Score/ScorePie.tsx
```diff
@@ -30,28 +43,33 @@ export const ScorePie = ({ pieDiameter, arguableId, arguableType }: Props) => {
...
+    <StyledBox display="flex" width={pieDiameter} height={pieDiameter}>
       <PieChart
         data={data}
-        radius={50} // is relative to viewbox size [100, 100]
+        radius={45} // is relative to viewbox size [100, 100]
         label={({ dataEntry }) => dataEntry.key}
         labelStyle={() => ({
           fontSize: "14", // is relative to viewbox size [100, 100]
           pointerEvents: "none",
+          webKitUserSelect: "none" /* Safari */,
```
keyserj, line 54: hmm I'm looking at the css in the browser and I don't see the webkit css prefix. not sure what the issue is here, BUT our styling solution (emotion) actually automatically adds prefixes - it [uses stylis by default](https://emotion.sh/docs/@emotion/cache#stylisplugins) and you can check out what should be generated by typing css into stylis's [homepage](https://stylis.js.org/). in this case, i think you should be able to do something like
```
labelStyle={() => css({ [insert css here] ... })
```
and it should add the prefixes (see emotion docs for [css object styling](https://emotion.sh/docs/object-styles#with-css))

web/src/modules/topic/components/Score/ScorePie.tsx
```diff
@@ -30,28 +43,33 @@ export const ScorePie = ({ pieDiameter, arguableId, arguableType }: Props) => {
...
+          msUserSelect: "none" /* IE 10 and IE 11 */,
+          userSelect: "none" /* Standard syntax */,
         })}
         segmentsStyle={{
+          transformOrigin: "center",
```
keyserj, line 59: was this necessary? mozilla says this is default https://developer.mozilla.org/en-US/docs/Web/CSS/transform-origin

web/src/modules/topic/components/Score/ScorePie.tsx
```diff
@@ -30,28 +43,33 @@ export const ScorePie = ({ pieDiameter, arguableId, arguableType }: Props) => {
...
           transition: theme.transitions.create("all", {
             duration: theme.transitions.duration.shortest,
           }),
+          pointerEvents: "visiblePainted",
```
keyserj, line 63: do you know if `visiblePainted` does anything special over `auto`? from a brief glance https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events it looks like they should behave the same for SVGs - and if that's the case, I'd just use `auto` here since that's equivalent to having no pointer-events property specified

keyserj (commented):
bunch of NP's - nice work figuring out that pointer-event jazz

web/src/modules/topic/components/Score/ScorePie.tsx
```diff
@@ -30,28 +43,33 @@ export const ScorePie = ({ pieDiameter, arguableId, arguableType }: Props) => {
...
         data={data}
-        radius={50} // is relative to viewbox size [100, 100]
+        radius={45} // is relative to viewbox size [100, 100]
         label={({ dataEntry }) => dataEntry.key}
         labelStyle={() => ({
           fontSize: "14", // is relative to viewbox size [100, 100]
           pointerEvents: "none",
+          webKitUserSelect: "none" /* Safari */,
+          msUserSelect: "none" /* IE 10 and IE 11 */,
+          userSelect: "none" /* Standard syntax */,
```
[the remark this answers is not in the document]
Nezzely: The label was always selectable, i just didn't think a user ever wanted to copy the text

web/src/modules/topic/components/Score/ScorePie.tsx
Nezzely: Its a habit for me to put other browser CSS along with the default. I can remove it if you want! (I have no preference)

web/src/modules/topic/components/Score/ScorePie.tsx
Nezzely: `pointerEvents: "auto"` will work just fine.

web/src/modules/topic/components/Score/ScorePie.tsx
Nezzely: Nice call out! i wasn't familiar with the syntax `labelStyle={() => css({ [insert css here] ... })`
works like a charm <3
`transform: dataIndex === hovered ? "scale(1.1)" : "",`

web/src/modules/topic/components/Score/ScorePie.tsx
```diff
@@ -18,6 +19,18 @@ export const ScorePie = ({ pieDiameter, arguableId, arguableType }: Props) => {
   const theme = useTheme();
   const [hovered, setHovered] = useState<number | undefined>(undefined);
 
+  const segmentMouseEvent = (event: React.MouseEvent, index: number | undefined) => {
+    const target = event.target as HTMLElement;
+    if (index === undefined) {
```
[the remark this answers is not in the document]
Nezzely: function removed in favor of segmentStyling

keyserj (approved):
steph-curry-halfcourt-shot

PR #107, opened 2023-03-25
Improve score pie uiux

alextilot:
### Description of changes
- added "buffer" circle behind score pie; catches segment rendering edge cases
- added conditional overflow hidden on body when workspace is rendered.


keyserj (approved):
Nice
flip-desk

PR #108, opened 2023-03-25
fix: criteria diagram correctly overlays table

alextilot:
### Description of changes
- criteria diagram wasn't correctly overlaying the criteria table, zIndex was changed.

### DIFF
--- a/web/src/pages/index.page.tsx
+++ b/web/src/pages/index.page.tsx
@@ -47,7 +47,8 @@ const Page = () => {
         </Box>
 
         {claimDiagramId && (
-          <Box width="100%" height="100%" position="absolute">
+          // Criteria Table has header (z-index:2); expectation: overlay the component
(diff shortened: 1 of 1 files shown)


web/src/pages/index.page.tsx
```diff
@@ -47,7 +47,7 @@ const Page = () => {
         </Box>
 
         {claimDiagramId && (
-          <Box width="100%" height="100%" position="absolute">
+          <Box width="100%" height="100%" position="absolute" zIndex="2">
```
keyserj, line 50: can we add a comment why we're using 2 here? "table toolbar and pinned headers have z-index set, and we want this in front"

alextilot: sure i suppose

keyserj (approved):
stand-aside

PR #115, opened 2023-04-12
Responsive font size rebase

### LINKED ISSUE #74: Fit more columns in the criteria table
The criteria table isn't easily usable with only two columns visible, and that is the maximum number of columns for a screen resolution width of 1280 px.

Allow 4 columns to fit in the criteria table for a screen resolution of 1280px. Some ideas for how this might be done:
- decrease the node size (and everything in it) to ~3/4 what it currently is (size was originally somewhat arbitrary because react-flow can zoom)
- remove bottom Y div and side X divs
image
- make node more of a square by allowing more textarea rows but narrowing the node width
- rotate column headers 45 deg (this would be sweet but seems technically hard, and editing should be done horizontally...)
image

- don't reuse the node component for cells, just put the text in with the score & claim indicator, then the cells can be responsive as material UI table intended

**Additional context**

**Technical ideas**

alextilot:
Closes #74

### Description of changes
- Adjusted  various web components to scale with root:font-size  when applicable

### Additional context
-  The user may change their default web browser font, EG: chrome://settings/fonts, we should respect their choice, and conform.

### DIFF
--- a/web/src/modules/topic/components/Node/EditableNode.styles.tsx
+++ b/web/src/modules/topic/components/Node/EditableNode.styles.tsx
@@ -1,8 +1,12 @@
 import styled from "@emotion/styled";
 import { TextareaAutosize } from "@mui/material";
 
-export const nodeWidth = 300;
-export const nodeHeight = 84;
+import { htmlDefaultFontSize } from "../../../../pages/_document.page";
+
+const nodeWidthRem = 16;
+
+export const nodeWidthPx = nodeWidthRem * htmlDefaultFontSize;
+export const nodeHeightPx = 66;
 
 export const YEdgeDiv = styled.div`
   display: flex;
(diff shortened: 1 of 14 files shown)


web/src/modules/topic/components/Node/EditableNode.styles.tsx
```diff
@@ -62,7 +64,7 @@ const options = {
 
 /* some copied from https://github.com/wbkd/react-flow/blob/147656b22f577bb4141664d000e62ada9b490473/src/theme-default.css#L42-L77 */
 export const NodeDiv = styled("div", options)<NodeDivProps>`
-  width: ${nodeWidth}px;
+  width: clamp(200px, 16rem, ${nodeWidth}px);
```
keyserj, line 67: unfortunately the table still doesn't fit 4 columns at 1280px screen width. but I think our node width can be significantly decreased - it's 16rem right now, but I think ~11rem still looks good
image

(I think `EditableNode#NodeDiv width` and `NodeCell#StyledDiv min-width` both need to be adjusted)

alextilot:
Created a new PR for these changes #117, closing this one.

PR #116, opened 2023-04-14
Resizing editable node

alextilot:
Partially Closes issue #74

### Description of changes
-  Topic Pane Drawer correctly sits inside the workspace height space. It no longer over flows the top.
-  Editable node has been tidied up, extra space around text area has been removed, thus increasing the text area's available space
-  Criteria Table gained a box shadow effect on the bottom; this gives WOW factor.

### Additional context
- This should be merged into master before the REM unit changes.

### DIFF
--- a/web/src/modules/topic/components/Node/EditableNode.styles.tsx
+++ b/web/src/modules/topic/components/Node/EditableNode.styles.tsx
@@ -1,8 +1,8 @@
 import styled from "@emotion/styled";
 import { TextareaAutosize } from "@mui/material";
 
-export const nodeWidth = 300;
-export const nodeHeight = 84;
+export const nodeWidth = 176;
+export const nodeHeight = 66;
 
 export const YEdgeDiv = styled.div`
   display: flex;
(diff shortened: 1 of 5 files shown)


keyserj (commented):
Thanks for splitting this out

Regarding the commit messages:
1. the phrasing should be in command form - you're telling the commit what it's doing. `feat: topic pane drawer is set to workspace height` -> `feat: set topic pane drawer to workspace height` (though saying "set height" makes it sound like there's a visual difference - I think rephrasing might be more accurate like `refactor(drawer): position relatively` then in extra description add: `instead of fixed`)
2. i'd generally reserve `feat` for more impactful changes. i'd call the drawer change a refactor and the other two `touchup`s

web/src/modules/topic/components/Node/EditableNode.styles.tsx
```diff
@@ -1,8 +1,8 @@
 import styled from "@emotion/styled";
 import { TextareaAutosize } from "@mui/material";
 
-export const nodeWidth = 300;
-export const nodeHeight = 84;
+export const nodeWidth = 256;
+export const nodeHeight = 66;
```
keyserj, line 5: what do you think about going even smaller? https://github.com/amelioro/ameliorate/pull/115#discussion_r1165817522

keyserj (approved):
like-a-glove

PR #117, opened 2023-04-17
Convert to rem units

alextilot:
### Description of changes
- Changed various components to scale with rem units: Topic Drawer, Criteria Table, Editable Node, Score Button
- Added documentation on sizing fonts and components

### DIFF
--- a/web/docs/uxui-guidelines.md
+++ b/web/docs/uxui-guidelines.md
@@ -1,0 +1,46 @@
+# UX / UI Guidelines
+
+## Table of Contents
+
+[Introduction](#Introduction)  
+[8pt grid system](#8pt-grid-system)  
+[Pixels and Accessibility](#pixels-and-accessibility)
+
+## Introduction
+
+Documentation for sharing UX/UI guidelines for the ameliorate project.
+
+## 8pt grid system
+
+These blog posts "[Everything you should know about 8 point grid system in UX design](https://uxplanet.org/everything-you-should-know-about-8-point-grid-system-in-ux-design-b69cb945b18d?gi=d321e6edb7fd)" and "[The 8-Point Grid](https://spec.fm/specifics/8-pt-grid)" offer reasons for sticking to sizes in multiples of 8 - check them out, but the points that seem most relevant are summarized in the below sections.
+
+### How to use it
+
+- Generally try to use px's that are multiples of 8 (e.g. 8, 16, 24, ...). For the case of rems, these will be multiples of 0.5.
+- Font size is an exception, use what looks nice.
+- For narrow spacings (padding, margin, line height), multiples of 4 can also work.
+- Treat this as a default, but if something looks better without meeting the guideline, just go with that.
+
+### Why?
+
+- It provides a visual hierarchy to elements and drives consistent scalability with fewer decisions.
+- Helps maintain a consistent feel across the website.
+- Helps designers and developers work faster on a project.
+- Various device screen sizes are multiples of 8.
+- Helps build responsive websites for various devices.
+
+## Pixels and Accessibility
+
+Using `rem` scales sizes with the user's default browser font size. The blog post "[The Surprising Truth About Pixels and Accessibility](https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/)" describes different use cases for using `rem` units and `px` units for web components. This is an overview of said post.
+
+## How to use it
+
+- Generally: use `px` for aesthetics, use `rem` for visual clarity or functionality.
+- There is no cookie cutter method to "correctly" pick between them, just try your best.
+- It can help to test what the sizing looks like at different default font sizes.
+  - Change the defaults: [Chrome](https://support.microsoft.com/en-us/microsoft-edge/increase-default-text-size-in-microsoft-edge-c62f80af-381d-0716-25a3-c4856dd3806c), [FireFox](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use), [Safari](https://support.apple.com/en-gb/HT207209), [Edge](https://support.microsoft.com/en-us/microsoft-edge/increase-default-text-size-in-microsoft-edge-c62f80af-381d-0716-25a3-c4856dd3806c).
+
+### Why?
+
+- The main accessibility consideration when it comes to pixel-vs-em/rem is vision. We want people with limited vision to be able to comfortably read the sentences and paragraphs on our websites and web applications.
+- As a general rule, we should give the user as much control as possible, and we should never disable or block their settings from working. For this reason, it's very important to use a relative unit like rem for typography.
--- a/web/src/modules/topic/components/Node/EditableNode.styles.tsx
+++ b/web/src/modules/topic/components/Node/EditableNode.styles.tsx
@@ -1,8 +1,12 @@
 import styled from "@emotion/styled";
 import { TextareaAutosize } from "@mui/material";
 
-export const nodeWidth = 176;
-export const nodeHeight = 66;
+import { htmlDefaultFontSize } from "../../../../pages/_document.page";
+
+const nodeWidthRem = 11;
+
+export const nodeWidthPx = nodeWidthRem * htmlDefaultFontSize;
+export const nodeHeightPx = 66;
 
 export const YEdgeDiv = styled.div`
   display: flex;
@@ -17,7 +21,7 @@ export const NodeTypeDiv = styled.div`
 `;
 
 export const NodeTypeSpan = styled.span`
-  font-size: 14px;
+  font-size: 0.875rem;
   line-height: 1;
   padding-left: 4px;
 `;
@@ -49,7 +53,7 @@ export const StyledTextareaAutosize = styled(TextareaAutosize)<StyledTextareaPro
   align-self: center;
   background-color: ${({ color }) => color};
   width: 100%;
-  font-size: 16px;
+  font-size: 1rem;
   line-height: 1;
   font-family: inherit;
 `;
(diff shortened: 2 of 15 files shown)


web/src/modules/topic/components/Score/Score.tsx
```diff
@@ -34,10 +34,17 @@ export const Score = ({ arguableId, arguableType, score }: ScoreProps) => {
...
+  const mainButtonRef = useRef<HTMLButtonElement | null>(null);
 
-  const buttonLength = indicatorLength; //px
-  const buttonLengthScaled = buttonLength * zoomRatio; //px * zoom
+  useEffect(() => {
+    if (mainButtonRef.current?.clientHeight) {
+      setRefButtonHeight(mainButtonRef.current.clientHeight);
+    }
+  }, [mainButtonRef]);
+
```
alextilot, line 45: I'm using another state variable here so the client Height is never null.

I couldn't think of any better solution.

Ideas?

web/docs/sizing.md
```diff
@@ -0,0 +1,57 @@
...
+[8pt grid system](#8pt-grid-system)  
+[Pixels and Accessibility](#pixels-and-accessibility)
+
+## Introduction
+
+Documentation for sharing UX/UI design decisions for the ameliorate project.
+
+# 8pt grid system
+
+The 8pt Grid principle states that multiples of 8 (8, 16, 24, 32, 40, 48, 56, etc.) are used for the layout, dimension, pad, and margin of elements.
```
keyserj, line 13: A lot of this is taken directly from the first article. I think we should mention the article first, and mention that we're just summarizing. Also, I think we can summarize the `How to use it` even further
```suggestion
These two blog posts offer reasons for sticking to sizes in multiples of 8 - check them out, but the points that seem most relevant are summarized in the below sections.

### How to use it

- Generally try to use px's that are multiples of 8 (e.g. 8, 16, 24, ...). For the case of rems, these will be multiples of 0.5.
- Font size is an exception, use what looks nice.
- For narrow spacings (padding, margin, line height), multiples of 4 can also work.
- Treat this as a default, but if something looks better without meeting the guideline, just go with that.

### Why?
[I think this section is good as is]
```

web/docs/sizing.md
```diff
@@ -0,0 +1,57 @@
...
+- Various device screen sizes are multiples of 8.
+- Helps build responsive websites for various devices.
+
+Sources:
+
+- [Everything you should know about 8 point grid system in UX design](https://uxplanet.org/everything-you-should-know-about-8-point-grid-system-in-ux-design-b69cb945b18d?gi=d321e6edb7fd)
+- [What Is 8 Point Grid System In UX Design](https://fronty.com/what-is-8-point-grid-system-in-ux-design/)
+- [The 8-Point Grid](https://spec.fm/specifics/8-pt-grid)
+
+# Pixels and Accessibility
```
keyserj, line 41: I think we should do the same thing as above - mention the source first, and say here is a summary of what seems most relevant

web/docs/sizing.md
```diff
@@ -0,0 +1,57 @@
...
+
+## How to use it
+
+- Use `rem` scales with the user's default browser font size.
+- If converting it to `rem` adds visual clarity or functionality, do it. Otherwise keep it fixed.
+- If you are unsure, test it. Change the defaults: [Chrome](https://support.microsoft.com/en-us/microsoft-edge/increase-default-text-size-in-microsoft-edge-c62f80af-381d-0716-25a3-c4856dd3806c), [FireFox](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use), [Safari](https://support.apple.com/en-gb/HT207209), [Edge](https://support.microsoft.com/en-us/microsoft-edge/increase-default-text-size-in-microsoft-edge-c62f80af-381d-0716-25a3-c4856dd3806c).
+- There is no cookie cutter method, it will take some intuition.
```
keyserj, line 48:
```suggestion
# Pixels and Accessibility

Using `rem` scales sizes with the user's default browser font size. [mention that we're summarizing from the source]

## How to use it

- Generally: use `px` for aesthetics, use `rem` for visual clarity or functionality.
- There is no cookie cutter method to "correctly" pick between them, just try your best.
- It can help to test what the sizing looks like at different default font sizes. Change the defaults: [Chrome](https://support.microsoft.com/en-us/microsoft-edge/increase-default-text-size-in-microsoft-edge-c62f80af-381d-0716-25a3-c4856dd3806c), [FireFox](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use), [Safari](https://support.apple.com/en-gb/HT207209), [Edge](https://support.microsoft.com/en-us/microsoft-edge/increase-default-text-size-in-microsoft-edge-c62f80af-381d-0716-25a3-c4856dd3806c).
```

keyserj (commented):
Nice, clean work. Commits are better, I'd just say that these are not `feat` changes, they're `refactor`s (and the last commit should be reworded if we do the `sizing.md`->`ux-guidelines.md` change)

web/docs/sizing.md
```diff
@@ -0,0 +1,57 @@
+##### Table of Contents
```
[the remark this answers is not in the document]
alextilot: Great idea, i added an anchor to direct users to the document via click

web/docs/sizing.md
```diff
@@ -0,0 +1,57 @@
...
+
+[Introduction](#Introduction)  
+[8pt grid system](#8pt-grid-system)  
+[Pixels and Accessibility](#pixels-and-accessibility)
+
+## Introduction
+
+Documentation for sharing UX/UI design decisions for the ameliorate project.
```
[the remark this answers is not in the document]
alextilot: I agree, good name

web/docs/sizing.md
```diff
@@ -0,0 +1,57 @@
...
+- It provides a visual hierarchy to elements and drives consistent scalability with fewer decisions.
+- Helps maintain a consistent feel across the website.
+- Helps designers and developers work faster on a project.
+- Various device screen sizes are multiples of 8.
+- Helps build responsive websites for various devices.
+
+Sources:
+
+- [Everything you should know about 8 point grid system in UX design](https://uxplanet.org/everything-you-should-know-about-8-point-grid-system-in-ux-design-b69cb945b18d?gi=d321e6edb7fd)
+- [What Is 8 Point Grid System In UX Design](https://fronty.com/what-is-8-point-grid-system-in-ux-design/)
```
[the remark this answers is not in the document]
alextilot: i thought it was a quick summarization so i added it, in hindsight we are already summarizing it. ill remove it.

web/src/modules/topic/components/Node/EditableNode.styles.tsx
```diff
@@ -64,7 +68,7 @@ const options = {
 
 /* some copied from https://github.com/wbkd/react-flow/blob/147656b22f577bb4141664d000e62ada9b490473/src/theme-default.css#L42-L77 */
 export const NodeDiv = styled("div", options)<NodeDivProps>`
-  width: ${nodeWidth}px;
+  width: clamp(200px, 11rem, ${nodeWidthPx}px);
```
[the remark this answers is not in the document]
alextilot: agree

web/src/modules/topic/components/Node/EditableNode.styles.tsx
```diff
@@ -64,7 +68,7 @@ const options = {
 
 /* some copied from https://github.com/wbkd/react-flow/blob/147656b22f577bb4141664d000e62ada9b490473/src/theme-default.css#L42-L77 */
 export const NodeDiv = styled("div", options)<NodeDivProps>`
-  width: ${nodeWidth}px;
+  width: clamp(200px, ${nodeWidthRem}rem, ${nodeWidthPx}px);
```
keyserj, line 71: nodeWidthRem = nodeWidthPx at 16 px default font size, so this clamp is actually making the width not scale for bigger default font sizes. I believe this is the reason for the very awkward problem node in the flow diagram
image

additionally, the 200px minimum is smaller than nodeWidthRem * 16 = 176px, so we're clamping even at the default font size, making the flow nodes wider than they should be

Just to be sure - are you intentionally clamping this? The buttons & font size within the node are not clamped in either direction, so the size of the node itself does not remain consistent with these outside of the bounds

web/src/modules/topic/utils/layout.ts
```diff
@@ -20,12 +20,10 @@ export const layout = async (nodes: Node[], edges: Edge[], orientation: Orientat
     "elk.layered.spacing.nodeNodeBetweenLayers": orientation === "DOWN" ? "90" : "135",
```
keyserj, line 20: this also needs to be updated. i played with some numbers and this is what I think looks decent at default font size 16. pls replace the existing `// not sure why larger...` comment on the line before with the comment I added here, so that it's clear that the 90 isn't only correlated to the node height (I thought it was at first)
```suggestion
    // these spacings are just what roughly seem to look good
    "elk.layered.spacing.nodeNodeBetweenLayers": orientation === "DOWN" ? "90" : "135",
    "elk.spacing.nodeNode": orientation === "DOWN" ? "20" : "50",
```

web/docs/uxui-guidelines.md
```diff
@@ -0,0 +1,55 @@
...
+[8pt grid system](#8pt-grid-system)  
+[Pixels and Accessibility](#pixels-and-accessibility)
+
+## Introduction
+
+Documentation for sharing UX/UI guidelines for the ameliorate project.
+
+## 8pt grid system
+
+These two blog posts offer reasons for sticking to sizes in multiples of 8 - check them out, but the points that seem most relevant are summarized in the below sections.
```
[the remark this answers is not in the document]
alextilot: That makes sense to me, we don't have lot of sources. having them inline would be nice.

web/src/modules/topic/components/Node/EditableNode.styles.tsx
alextilot: `nodeWidthRem = nodeWidthPx at 16 px default font size, so this clamp is actually making the width not scale for bigger default font sizes. I believe this is the reason for the very awkward problem node in the flow diagram`
That is correct. I totally  missed the problem node having 3 buttons.

I can add a clamp to the buttons and text to fit the node at maxWidth, and not allow additional scaling.

the 200px was intentional, I can set it to the browser Default min

keyserj (approved):
Wooooot nice work. It's good that we're getting this set up sooner than later - I probably wouldn't've gotten to it for a while 😆  
shrek-rats-see-clearly

PR #151, opened 2023-05-14
Feature: Adde tooltip for view criteria table and view claims buttons

### LINKED ISSUE #147: Add tooltip for "view criteria table" and "view claims" buttons
The toolbar buttons and add node buttons already have tooltips, but the node indicator buttons do not yet. We should add these.
image

Something like this (Material UI Tooltip)
image

or like this (html `title` and `aria-label`)
image

We should probably be consistent with these but I'm not actually sure which one is better. Maybe Material UI Tooltip because it must meet Material UI standards?

**Additional context**

**Technical ideas**

tarunsamanta2k20:
Closes #147 

### Description of changes
The toolbar buttons and add node buttons already have tooltips, but the node indicator buttons do not yet. We should add these.
- 

### Additional context
image
image

-

### DIFF
--- a/web/src/modules/topic/components/Indicator/ClaimIndicator.tsx
+++ b/web/src/modules/topic/components/Indicator/ClaimIndicator.tsx
@@ -1,5 +1,4 @@
 import { Article, ArticleOutlined } from "@mui/icons-material";
-import Tooltip from "@mui/material/Tooltip";
 
 import { useExplicitClaimCount } from "../../store/arguableHooks";
 import { viewOrCreateClaimDiagram } from "../../store/viewActions";
@@ -17,17 +16,14 @@ export const ClaimIndicator = ({ arguableId, arguableType }: Props) => {
   const Icon = explicitClaimCount > 0 ? Article : ArticleOutlined;
 
   return (
-
-    <Tooltip title="View claims">
-
-    <Indicator  Icon={Icon} onClick={(event) => {
+    <Indicator
+      Icon={Icon}
+      title={"View claims"}
+      onClick={(event) => {
         // prevent setting the node as selected because we're about to navigate away from this diagram
         event.stopPropagation();
         viewOrCreateClaimDiagram(arguableId, arguableType);
-      }}  />
-        
-    </Tooltip>
-
-   
+      }}
+    />
   );
 };
--- a/web/src/modules/topic/components/Indicator/ClaimIndicator.tsx
+++ b/web/src/modules/topic/components/Indicator/ClaimIndicator.tsx
@@ -1,4 +1,5 @@
 import { Article, ArticleOutlined } from "@mui/icons-material";
+import Tooltip from "@mui/material/Tooltip";
 
 import { useExplicitClaimCount } from "../../store/arguableHooks";
 import { viewOrCreateClaimDiagram } from "../../store/viewActions";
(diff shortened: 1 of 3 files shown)


web/src/modules/topic/components/Indicator/Indicator.tsx
```diff
@@ -8,40 +8,48 @@ interface IndicatorProps {
...
-    // this follows the MUI example, but I can't actually see the hover at all
-    // <StyledIconButton onClick={onClick} disabled={!onClick}>
-    //   <Icon color="neutralContrast" />
-    // </StyledIconButton>
-
-    // no hover effect (this isn't even a button)
-    // <Icon />
-  );
-};
+export const Indicator = forwardRef<HTMLButtonElement, IndicatorProps>(
```
keyserj, line 11: What do you think about putting the `<Tooltip>` in this component directly, and accepting something like a `tooltipText` prop? I like how `forwardRef` allows our parent component to decide whether or not it wants to add a tooltip, but I think it's likely that all indicators will want a tooltip

web/src/modules/topic/components/Indicator/ClaimIndicator.tsx
```diff
@@ -16,13 +17,17 @@ export const ClaimIndicator = ({ arguableId, arguableType }: Props) => {
   const Icon = explicitClaimCount > 0 ? Article : ArticleOutlined;
 
   return (
-    <Indicator
-      Icon={Icon}
-      onClick={(event) => {
+
+    <Tooltip title="View claims">
+
+    <Indicator  Icon={Icon} onClick={(event) => {
```
keyserj, line 23: I'm noticing the spacing is off - I think prettier needs to be run `npx prettier --write .`. I can handle this if need be.

I realize the contributing guide doesn't mention prettier, nor is it set up automatically for contributors - I think I'll add a precommit hook and/or github action to have this run

tarunsamanta2k20:
@keyserj the no need to pass forwardRef. I can simply move Tooltip there and pass Tooltip text as a prop. what would you think?

web/src/modules/topic/components/Indicator/CriteriaTableIndicator.tsx
```diff
@@ -21,5 +22,11 @@ export const CriteriaTableIndicator = ({ nodeId, diagramId }: Props) => {
 
   const Icon = hasCriteria ? TableChart : TableChartOutlined;
 
-  return <Indicator Icon={Icon} onClick={() => viewCriteriaTable(nodeId)} />;
+  return (
+    <Indicator
+      Icon={Icon}
+      title={"view criteria table"}
```
keyserj, line 28:
```suggestion
      title={"View criteria table"}
```

web/src/modules/topic/components/Indicator/CriteriaTableIndicator.tsx
```diff
@@ -1,4 +1,5 @@
 import { TableChart, TableChartOutlined } from "@mui/icons-material";
+import React from "react";
```
keyserj, line 2: no longer being used
```suggestion
```

keyserj (commented):
Nice work on this, just two np's

keyserj:
@all-contributors please add @tarunsamanta2k20 for code

PR #165, opened 2023-05-19
feat: added carousel with readme images to home page

### LINKED ISSUE #146: Add demo images to home page in carousel
We've got a nice demo video on the home page to showcase the tool, but we should throw some pictures in there too, so that people can get the idea even if they don't want to watch a video

Using a carousel seems good. Don't auto switch slides. Might need to play with whatever carousel options there are to see what looks nice
image

As for which images to use - I think the three at the [top of the readme](https://github.com/amelioro/ameliorate#ameliorate) would be good

**Additional context**

**Technical ideas**
[react-material-ui-carousel](https://github.com/Learus/react-material-ui-carousel) looks like a good carousel component

kaiseradr:
Closes #146

### Description of changes
* added demo images to home page in carousel


web/src/pages/index.page.tsx
```diff
@@ -50,8 +54,30 @@ const Home: NextPage = () => {
 
       <section id="how-it-works">
         <StyledBox>
-          <Typography variant="h4">How it works</Typography>
-          <YoutubeEmbed embedId="yM8RrwQWeJc" />
+          <Box sx={{ height: 600, width: "70%" }}>
+            <Typography variant="h4">How it works</Typography>
+            <Carousel autoPlay={false}>
```
keyserj, line 59: It's a little jarring when the carousel changes size due to its contents changing. I think it'd be ideal to have its height be set to the largest child's height, then center the children within the carousel if they're smaller than the carousel, but I'm not sure how to accomplish this simply while allowing the carousel's size to also be responsive based on screen size.

I think it's fine to leave this issue unsolved for now.

keyserj, line 59: Thoughts on making the nav buttons always visible? I think it'd be nicer for switching around

web/src/pages/index.page.tsx
```diff
@@ -50,8 +54,30 @@ const Home: NextPage = () => {
 
       <section id="how-it-works">
         <StyledBox>
-          <Typography variant="h4">How it works</Typography>
-          <YoutubeEmbed embedId="yM8RrwQWeJc" />
+          <Box sx={{ height: 600, width: "70%" }}>
```
keyserj, line 57: Do we need this div? Seems to me that any styling here can be put directly on the Carousel (by using emotion to make a `StyledCarousel`, like [this](https://github.com/amelioro/ameliorate/blob/1042b9615a55f69e02cc3a7d8805a977c2ad00a8/web/src/pages/index.style.tsx#L4-L10))

keyserj: Regarding the width - 70% is really small for smaller screens, but I do like having some margin. Maybe 90%?

web/src/pages/index.page.tsx
```diff
@@ -50,8 +54,30 @@ const Home: NextPage = () => {
...
+            <Typography variant="h4">How it works</Typography>
+            <Carousel autoPlay={false}>
+              <YoutubeEmbed embedId="yM8RrwQWeJc" />
```
keyserj, line 60: Since the `YoutubeEmbed` has a max width smaller than the Carousel's, I think we should horizontally center children of the Carousel. Maybe `flex` with `align-items` on a `styled(Carousel)` would work?
image

keyserj (commented):
Lookin good so far, couple preliminary comments. Since it's in draft, can you let me know when you'd like another review by re-requesting review?

Comments separate from the code:
- the PR description's issue number should replace the [insert issue number], like `#146`. `#` is how you can link to other GitHub items like issues or PRs
- this isn't a problem, but I see that you've named your fork with a suffix `-forked`. standard practice is to just keep the same repo name - it'll be distinguished by the account that it's under (your account, instead of amelioro)

web/src/pages/index.page.tsx
```diff
@@ -1,8 +1,12 @@
 import { ArrowDownward } from "@mui/icons-material";
 import { Box, Button, Divider, Typography } from "@mui/material";
+// import Image from "material-ui-image";
+// Cannot find module '@material-ui/core/CircularProgress' --> error I get when I try to use the above package --> does it really have dependency on CircularProgress
```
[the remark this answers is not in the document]
kaiseradr: Just a general comment on this commit - I was experimenting with a few different things and didn't think you'd look at it until I marked it ready for review. So that probably explains the unfinished/confusing state of this. Nonetheless, I appreciate the feedback. :)

I was messing around with `material-ui-image` before I saw a ESLint warning suggesting that I use Next's Image component.

web/src/pages/index.page.tsx
kaiseradr: Yeah, I was annoyed by this behavior as well, but didn't find a good solution yet.

web/src/pages/index.page.tsx
kaiseradr: It's hard to remember, but this may have been an attempt to lock in a consistent height for all images- I think it was just experimental. Good point on using a `StyledCarousel` for formatting.

web/src/pages/index.page.tsx
kaiseradr: I think that's a good call.

web/src/pages/index.page.tsx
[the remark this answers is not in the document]
keyserj: allowing the base URL via next.config.js should be fine. storing the image in the repo will not work if the image is trying to be retrieved from the client (if you look at other images currently stored in the repo, they're stored under `public/`, so they're exposed to the web and can be accessed from the client)

web/src/pages/index.page.tsx
keyserj: generally `styled` is preferred over `sx` and `style` because
- it doesn't muddle up or mix with the html
- we can give meaningful names to the styled component (though we don't always do this)
- it applies css via class and therefore
  - allows control over related-element (e.g. child, sibling) css
  - allows overriding css
  - makes it easy to reuse css

`sx`/`style` can be fine if there are only a few, obviously one-off rules to apply. I'll mainly use them during prototyping if I haven't made a `styled` component yet, and only a few rules end up being used, I might just leave it partially out of laziness.

web/src/pages/index.page.tsx
kaiseradr: maybe prioritize maintaining aspect ratio for now...

kaiseradr:
@keyserj Just added a commit. I managed to fix a few issues but also had to hard-code some things, which is not ideal. It's not working the way I want, but I think there is some progress. Let's a take a look together over discord sometime today if you have time and hash out issues there instead of through comments if you're good with that.

keyserj:
Yep you can just ping me on discord

web/src/pages/index.page.tsx
```diff
@@ -51,7 +55,12 @@ const Home: NextPage = () => {
       <section id="how-it-works">
         <StyledBox>
           <Typography variant="h4">How it works</Typography>
-          <YoutubeEmbed embedId="yM8RrwQWeJc" />
+          <StyledCarousel autoPlay={false} navButtonsAlwaysVisible={true}>
+            <YoutubeEmbed embedId="yM8RrwQWeJc" />
+            <Image src={DemoPicture1} alt="mapping solutions to problems" />
+            <Image src={DemoPicture2} alt="comparing solutions" />
+            <Image src={DemoPicture3} alt="justifying and critiquing claims" />
```
kaiseradr, line 62: I wasn't sure how we wanted to name the images, so I just tried to summarize the contents for the `alt` property and give generic names for imports. Let me know what you'd like to see here.

On a separate note, I will squash all the commits into one after you approve the PR and before I merge it (if that's what you're looking for).

keyserj: 1. re: naming; you should name the imported variables similarly to how you've named the image files, except use camelCase. E.g. comparingSolutions
2. I actually overlooked importing images and using them this way 😅 I think the way you've done it is better than [what I'm doing elsewhere](https://github.com/amelioro/ameliorate/blob/4ca37e071b884e3116414ccea6c90920ae6cfa0a/web/src/common/components/Layout.tsx#L29-L34). After reading the docs more closely, I'm pretty sure this way I'm doing elsewhere is actually treating the images as remote, but I didn't have to add them to the `remotePatterns` whitelist most likely because they're remotely accessed from the webserver itself. And since importing is the proper way to locally retrieve an image, I'm pretty sure nextjs will handle sending it to the client without needing to host it via the `public` directory
3. go ahead and squash the commits after you've addressed my suggestions. Until you're a member of the repository, you can't merge PRs yourself

web/src/pages/index.style.tsx
```diff
@@ -8,3 +9,9 @@ export const StyledBox = styled(Box)`
   margin: 5rem 0;
   text-align: center;
 `;
+
+export const StyledCarousel = styled(Carousel)`
+  width: 90%;
+  height: 100%;
+  max-width: 1165px;
```
keyserj, line 16:
```suggestion
  max-width: 1165px; // roughly the size of the local images used in the carousel
```

keyserj:
side note: I see you have a merge commit in your commit history. Squashing should make this find, but FYI that generally I think fast-forwarding or rebasing changes from other PRs keeps PR commit history cleaner. See my notes here https://github.com/amelioro/ameliorate/pull/130#pullrequestreview-1404726806

web/src/pages/index.page.tsx
kaiseradr: 1. Will do.
2. I'm pretty sure I saw this in an example somewhere and decided to use it. Glad it was helpful. :)
3. Oops, that makes sense.

kaiseradr:
> side note: I see you have a merge commit in your commit history. Squashing should make this fine, but FYI that generally I think fast-forwarding or rebasing changes from other PRs keeps PR commit history cleaner. See my notes here [#130 (review)](https://github.com/amelioro/ameliorate/pull/130#pullrequestreview-1404726806)

web/src/pages/index.page.tsx
keyserj: regarding the danger of overwriting history - as long as you've committed changes at some point, you should be able to retrieve them via [git reflog](https://git-scm.com/docs/git-reflog). commits are immutable and will only be deleted during automatic git garbage collection if they're dangling (not pointed to by a branch or another commit) for ~2 weeks (if I understand the [docs](https://git-scm.com/docs/git-gc#Documentation/git-gc.txt-gcpruneExpire) correctly...).

I think this article https://blog.git-init.com/does-amending-a-commit-change-its-hash/ (and the previous one it links to) explains git commit storage really well.

still happy to pair on commit management if you want. feel free to ping me on discord

web/src/pages/index.page.tsx
```diff
@@ -51,7 +55,12 @@ const Home: NextPage = () => {
       <section id="how-it-works">
         <StyledBox>
           <Typography variant="h4">How it works</Typography>
-          <YoutubeEmbed embedId="yM8RrwQWeJc" />
+          <StyledCarousel autoPlay={false} navButtonsAlwaysVisible={true}>
+            <YoutubeEmbed embedId="yM8RrwQWeJc" />
+            <Image src={mappingSolutionsToProblems} alt="mapping solutions to problems" />
+            <Image src={comparingSolutions} alt="comparing solutions" />
+            <Image src={justifyingAndCritiquingClaims} alt="justifying and critiquing claims" />
```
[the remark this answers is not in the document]
kaiseradr: Good catch and thanks for finding the solution to this. I added another commit for these changes and then rebased again.

keyserj (approved):
Nice work on this :)
carousel

@all-contributors please add @kaiseradr for code

keyserj:
@all-contributors please add @kaiseradr for code

PR #166, opened 2023-05-19
build: lint commits on pr

### LINKED ISSUE #127: Add commit standard checks for Pull Requests
As a developer I want commit history to be checked on pull requests to keep the conventional commit standardized. 

Using Github Actions, check the commits and verify they are valid during a Pull Request.

Allowing users to self govern their commits

**Additional context**
[About status checks
](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks)

[Creating CI tests with the Checks API](https://docs.github.com/en/apps/creating-github-apps/guides/creating-ci-tests-with-the-checks-api)

**Technical ideas**

alextilot:
Closes #127 

### Description of changes
- updated the CI process to check commit messages for correctness. 
- using conventional commits as the standard.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,7 @@ This project uses commit hooks to automate some tasks, and these are managed via
 
 #### Conventional commits
 
-[Conventional commits](https://www.conventionalcommits.org/) is a standard format for commit messages, used to aid readability of commit history. This commit message format, with [these commit types](https://github.com/amelioro/ameliorate/blob/main/commitlint.config.js), is enforced in a commit hook via [commitlint](https://github.com/conventional-changelog/commitlint).
+[Conventional commits](https://www.conventionalcommits.org/) is a standard format for commit messages, used to aid readability of commit history. The format is `<type>[optional scope]: <description>` and an example commit message looks like `feat(header): add link to feedback page`. This commit message format, with [these commit types](https://github.com/amelioro/ameliorate/blob/main/commitlint.config.js), is enforced in a commit hook and github action via [commitlint](https://github.com/conventional-changelog/commitlint).
 
 #### Prettier
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,13 @@ name: ci
 on: pull_request
 
 jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: wagoid/commitlint-github-action@v5
   lint:
     runs-on: ubuntu-latest
     defaults:
@@ -17,3 +24,4 @@ jobs:
         if: always() # get formatting errors even if there are errors above
       - run: npm run lint
         if: always() # get lint errors even if there are errors above
+
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -1,5 +1,7 @@
 module.exports = {
   extends: ["@commitlint/config-conventional"],
+  helpUrl:
+    "https://github.com/amelioro/ameliorate/blob/main/CONTRIBUTING.md#conventional-commits",
   rules: {
     "type-enum": [
       2,
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -6,20 +6,19 @@ module.exports = {
     "type-enum": [
       2,
       "always",
-      // custom types should have comments explaining them
       [
-        "build",
-        "chore",
-        "ci",
-        "docs",
-        "feat",
-        "fix",
+        "build", // modifies the build process / pipeline for the application
+        "chore", // is a catch-all for any change that doesn't fit into another category
+        "ci", // modifies the continuous integration pipeline
+        "docs", // adds or improves documentation
+        "feat", // adds new functionality
+        "fix", // fixes a bug
         "touchup", // small improvement, too small for "feat" or "refactor"
-        "perf",
-        "refactor",
-        "revert",
-        "style",
-        "test",
+        "perf", // improves performance without functionality changes
+        "refactor", // doesn't add functionality and doesn't fix a bug
+        "revert", // undoes a previous commit
+        "style", // does not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
+        "test", // updates testing
       ],
     ],
   },


alextilot:
Couple things I'm on the fence about,
1) Should it be separated into its own file, or be a sub job in CI.yaml
2) If separated, how do you feel about the name / job name, i do not fancy it

3) commit scope, this is a part of continuous integration (ci) but it might get confused with file? idk.

.github/workflows/commitint.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+
+      - name: Install dependencies
+        run: npm install
```
keyserj, line 19: as is done in ci.yml, we should be able to use `npm ci` here instead of `npm install`. `npm ci` is intended for automated environments because it doesn't try to update dependencies, only install them as exactly appears in package-lock.json https://stackoverflow.com/a/53325242

.github/workflows/commitint.yml
```diff
@@ -0,0 +1,23 @@
...
+
+      - name: Validate PR commits with commitlint
+        if: github.event_name == 'pull_request'
+        run: npx commitlint --from ${{ github.event.pull_request.base.sha }} --to ${{ github.event.pull_request.head.sha }} --verbose
```
keyserj, line 23: funny story.... i think that commitlint action we saw should actually work for us - i found that it [accepts the commitlint.config.js](https://github.com/wagoid/commitlint-github-action/blob/master/action.yml#L5-L9), so it should use our custom commit types.

in any case, i advocate to keep your implementation until we run into a bug or issue that the repo action has solved. your implementation is much simpler. the only thing i currently see is that the repo action will output a [link to explain the convention](https://github.com/wagoid/commitlint-github-action/blob/master/action.yml#L25-L28), but i think maybe we can do this too if the messaging is unclear. i've got to see what the messaging looks like when it fails

.github/workflows/commitint.yml
```diff
@@ -0,0 +1,23 @@
+name: commitlint
```
keyserj, line 1: we can update CONTRIBUTING.md to say `is enforced in a commit hook and GitHub action via commitlint`

.github/workflows/commitint.yml
```diff
@@ -0,0 +1,23 @@
...
+          fetch-depth: 0
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+
+      - name: Install dependencies
+        run: npm install
+
+      - name: Validate PR commits with commitlint
+        if: github.event_name == 'pull_request'
```
keyserj, line 22: huh aren't we already only running this on `pull_request`?

.github/workflows/commitint.yml
keyserj, line 1: your commit type is `build(ci)` but this should just be a `ci` type. maybe `ci(commitlint)` to distinguish it from the `ci.yml` file. `build` is for building the project, like if we added a compile step or something.

keyserj, line 1: i actually think it'd probably be ever-so-slightly better to put this as a separate job in `ci.yml` because it is indeed another ci step. we can consider splitting it into a separate file if `ci.yml` gets to be big.

keyserj:
Also, can you post a screenshot of the action's results (messages if any) for the following cases?:
1. commit-1-good-message commit-2-good-message
2. commit-1-bad-message commit-2-good-message
3. commit-1-good-message commit-2-bad-message

to ensure we've tested the edges. otherwise i can also do this, just don't want to mess your commits if you're making changes

.github/workflows/commitint.yml
alextilot: perfect sounds good to me!

alextilot: I think ill make a commit template in the readme, should help some people out (me) I always forget what we have.

having links is nice for additional reading but i prefer a one stop shop

alextilot:
**Testing CI commit linting:**

**1. TEST: All commits are valid**
**Inputs:**
ci: integrate commitlint into github pr
docs(contributing): define conventional commits
docs(contributing): add table of contents

**Outputs (Expected-NoError):**
chrome_vDiX3rwEJD



**2. TEST: Last commit is invalid**
**Inputs:**
ci: integrate commitlint into github pr
docs(contributing): define conventional commits
docs(contributing): add table of contents
ERROR: error out

**Outputs (Expected-Error):**
chrome_wMxva0REVu



**2. TEST: Middle commit is invalid**
**Inputs:**
ci: integrate commitlint into github pr
docs(contributing): define conventional commits
docs(contributing): add table of contents
ERROR: error out
touchup: correct styling commit

**Outputs (Expected-Error):**
chrome_VSLCIO6f0h

CONTRIBUTING.md
```diff
@@ -6,6 +6,27 @@ All contributors (with GitHub accounts) will be recognized by being added to the
 
 If you're new to open source, you'll probably find [this open source guide](https://opensource.guide/how-to-contribute) useful.
 
+---
+
+## Table of Contents
```
keyserj, line 11: All markdown files automatically have a table of contents if you read them on GitHub https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/. Do you think it's still worth having this? Might be annoying to maintain

keyserj (commented):
Pretty heckin' baller. Just a couple readme suggestions

CONTRIBUTING.md
alextilot: That is fair, I never noticed that functionality. If that is the case I can drop the table of contents commit.

CONTRIBUTING.md
```diff
@@ -6,6 +6,27 @@ All contributors (with GitHub accounts) will be recognized by being added to the
...
+
+- [Code of Conduct](#code-of-conduct)
+- [Providing feedback](#providing-feedback)
+- [Finding an issue to work on](#finding-an-issue-to-work-on)
+- [Running the project](#running-the-project)
+- [Codebase overview](#codebase-overview)
+- [Working with the code](#working-with-the-code)
+  - [Code Style](#code-style)
+    - [Commit hooks](#commit-hooks)
```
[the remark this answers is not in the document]
alextilot: due to removal of the table of contents, this won't apply

keyserj (approved):
Woot nice work
lint-cat

PR #170, opened 2023-05-31
Set up backend with users and topics

keyserj:
### Description of changes

- Sets up backend structure (nextjs API with tRPC invoked via react query, postgres db, auth0 auth provider)
- Adds user login (behind hidden login button, can login via endpoint directly)
- Adds user topics (diagram data not yet associated to topics)

### Additional context

-


keyserj:
🎉

PR #209, opened 2023-08-28
163 hovering score logs anchorel prop is invalid error

### LINKED ISSUE #163: Hovering score logs "`anchorEl` prop is invalid" error
**Describe the bug**
In dev mode, when hovering a score for the first time after page load, an error is logged to the console:
```
Failed prop type: MUI: The `anchorEl` prop provided to the component is invalid.
It should be an HTML element instance or a virtualElement 
```
chrome_2023-05-18_14-43-21

**To Reproduce**
1. Go to solve page
2. Hover score
3. See error in console

**Expected behavior**
No error in console

**Screenshots**

**Additional context**
This bug isn't causing problems right now, because that we're using the [React Popper component](https://mui.com/material-ui/react-popper/) in a way that's unintended. Visual functionality should remain the same.

**Technical ideas**
We're using a [Popper](https://github.com/amelioro/ameliorate/blob/606ccc90a2f69a6a2178f3d9a7903fc748e51883/web/src/modules/topic/components/Score/Score.tsx#L80-L97) for the Score so that it can pop in front of everything else, but we're also using a [BackdropPopper](https://github.com/amelioro/ameliorate/blob/606ccc90a2f69a6a2178f3d9a7903fc748e51883/web/src/modules/topic/components/Score/Score.tsx#L63-L77) behind that so that we can ensure that the ScorePopper closes when hovering away (without the backdrop, if you hover the score button and move your mouse away before the ScorePopper shows in front, you could get the ScorePopper to stay open). The BackdropPopper is [stretched](https://github.com/amelioro/ameliorate/blob/606ccc90a2f69a6a2178f3d9a7903fc748e51883/web/src/modules/topic/components/Score/Score.styles.tsx#L43) so that hover into it will close the ScorePopper, and because of this, it didn't make sense for it to have an `anchorEl` set.

We should probably actually use a [Modal](https://mui.com/material-ui/react-modal/) instead of this BackdropPopper that behaves like a Modal.

keyserj:
Closes #163 

### Description of changes

- see commit messages

### Additional context

-

errors not displaying locally:
chrome_2023-08-28_16-31-52

### DIFF
--- a/src/web/topic/components/Node/AddNodeButton.tsx
+++ b/src/web/topic/components/Node/AddNodeButton.tsx
@@ -1,5 +1,3 @@
-import { Tooltip } from "@mui/material";
-
 import { addNode } from "../../store/createDeleteActions";
 import { type RelationDirection } from "../../utils/diagram";
 import { Relation } from "../../utils/edge";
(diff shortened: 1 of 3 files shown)


keyserj:
wow hey great work on this

PR #260, opened 2023-10-18
Add delete button to edges and nodes

### LINKED ISSUE #221: Button to delete selected node/edge
Right now the only way to delete a node/edge is by right-click, but that's not obvious.

Other diagramming tools (draw.io, miro) usually have a delete button in addition to a right-click menu option for deleting, so that seems like a good way to make the delete functionality visible.
image

**Additional context**
Discord convo https://discord.com/channels/1057707973482401892/1057707973482401899/1148994000909516870

**Technical ideas**
- Probably a "useSelectedGraphParts" hook to determine if the button is enabled, and to get the parts to delete if clicked
- Should be able to use the existing `deleteNode` and `deleteEdge` actions
- mui has a nice "delete" icon https://mui.com/material-ui/material-icons/?query=de&selected=Delete

seanyeh:
Closes #221 

What do you think about having a delete button as part of the node/edge? I hacked up a prototype to see what it would look like. I think it's more discoverable here than in the menu, as suggested in the linked issue. But totally understand if you think it's taking up valuable real estate.

Screenshot_2023-10-17_19-31-58

### Description of changes

- Added delete button to node/edge. (Is `Indicator` the right term for this? Wasn't sure what this was supposed to mean)

### Additional context

-

### DIFF
--- a/src/web/topic/components/Surface/TopicToolbar.tsx
+++ b/src/web/topic/components/Surface/TopicToolbar.tsx
@@ -1,4 +1,4 @@
-import { Build, Download, Error, Group, Redo, Undo, Upload } from "@mui/icons-material";
+import { Build, Delete, Download, Error, Group, Redo, Undo, Upload } from "@mui/icons-material";
 import { AppBar, Divider, IconButton, ToggleButton, Toolbar, Tooltip } from "@mui/material";
 import fileDownload from "js-file-download";
 import { useEffect, useState } from "react";
@@ -12,6 +12,8 @@ import {
   resetPerspectives,
   useIsComparingPerspectives,
 } from "../../../view/store/store";
+import { deleteGraphPart } from "../../store/createDeleteActions";
+import { useSelectedGraphPart } from "../../store/graphPartHooks";
 import { migrate } from "../../store/migrate";
 import { TopicStoreState, useOnPlayground } from "../../store/store";
 import { useUserCanEditTopicData } from "../../store/userHooks";
@@ -63,6 +65,8 @@ export const TopicToolbar = () => {
   const [canUndo, canRedo] = useTemporalHooks();
   const isComparingPerspectives = useIsComparingPerspectives();
   const [hasErrored, setHasErrored] = useState(false);
+
+  const selectedGraphPart = useSelectedGraphPart();
 
   const [isMoreActionsDrawerOpen, setIsMoreActionsDrawerOpen] = useState(false);
 
@@ -116,6 +120,22 @@ export const TopicToolbar = () => {
               disabled={!canRedo}
             >
               <Redo />
+            </IconButton>
+
+            <Divider orientation="vertical" />
+
+            <IconButton
+              color="inherit"
+              title="Delete"
+              aria-label="Delete"
+              onClick={() => {
+                if (selectedGraphPart) {
+                  void deleteGraphPart(selectedGraphPart);
+                }
+              }}
+              disabled={!selectedGraphPart}
+            >
+              <Delete />
             </IconButton>
           </>
         )}
--- a/src/web/topic/store/createDeleteActions.ts
+++ b/src/web/topic/store/createDeleteActions.ts
@@ -3,7 +3,8 @@ import { createDraft, finishDraft } from "immer";
 import { errorWithData } from "../../../common/errorHandling";
 import { emitter } from "../../common/event";
 import {
-  type Node,
+  type GraphPart,
+  Node,
   RelationDirection,
   buildEdge,
   buildNode,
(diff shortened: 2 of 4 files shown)


src/web/topic/components/Indicator/DeleteIndicator.tsx
```diff
@@ -0,0 +1,26 @@
...
+  graphPartId: string;
+  graphPartType: string;
+}
+
+export const DeleteIndicator = ({ graphPartId, graphPartType }: Props) => {
+  return (
+    <Indicator
+      Icon={Delete}
+      title={"Delete"}
+      onClick={ () => {
```
keyserj, line 16: prettier is angry here. there's a precommit hook to fix whitespace errors, but you can also use `npx prettier --write .` if you prefer

src/web/topic/components/Indicator/DeleteIndicator.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { deleteEdge, deleteNode } from "../../store/createDeleteActions";
+import { Indicator } from "./Indicator";
+
+interface Props {
+  graphPartId: string;
+  graphPartType: string;
+}
+
+export const DeleteIndicator = ({ graphPartId, graphPartType }: Props) => {
```
keyserj, line 11: if this is a user's topic, we'll want to restrict this functionality from those that aren't the topic's creator, like this https://github.com/amelioro/ameliorate/blob/9f2f1d9f571b51ecb14183c418e2a33eb3d6921d/src/web/topic/components/Surface/TopicToolbar.tsx#L87-L97

src/web/topic/store/utilActions.ts
```diff
@@ -37,6 +37,16 @@ export const setTopicData = (state: TopicStoreState, sessionUsername?: string) =
   emitter.emit("loadedTopicData", getTopicDiagram(useTopicStore.getState()));
 };
 
+export const getSelectedEdge = () => {
```
seanyeh, line 40: @keyserj is this sort of what you had in mind? as an aside what do you think about having some common interface for both nodes/edges in order to deduplicate some logic?

keyserj: This is the right logic, but I think a hook is more suitable than an action here because an action doesn't "watch" for changes to re-render, so the `disabled` check can become outdated when selection changes.

It seems to be working right now because there's another hook triggering a re-render when selection changes - I think it's [useTemporalHooks](https://github.com/amelioro/ameliorate/blob/c1458cc15e51e1e1bd15fc079df308f2d1f30fb7/src/web/topic/components/Surface/TopicToolbar.tsx#L63) since undo/redo currently include changes to selection (which is not ideal but is how it works right now ⚰️). It's fine now but if we were to extract undo/redo into a separate component, selection changes would no longer re-render the whole toolbar along with the `disabled` check.

(I also thought that hook logic would only run when the store changes, saving some performance over an action, but it appears that hook selectors run on all re-renders... I'll look into this more separately)

Secondly, I think using a common interface for nodes/edges is probably worthwhile, since I think there will be other actions that will apply to both nodes & edges.

So my thought is that we could make something like `useSelectedGraphPart` (or "parts" to be more future-proof when we can select multiple, but that might be more work). Probably would live in `graphPartHooks.ts`. Then when we use the graph part in the button's `onClick`, we could invoke a new `deleteGraphPart(graphPart)` action that checks if the graph part is a node/edge (probably easiest to just check [type === "ScoreEdge"](https://github.com/amelioro/ameliorate/blob/c1458cc15e51e1e1bd15fc079df308f2d1f30fb7/src/web/topic/utils/diagram.ts#L79) for now) and calls the specific delete method.

seanyeh: cool, TIL about hooks. I've made the changes and I've made a `GraphPart` interface for now. could consider having Node and Edge extend `GraphPart` (and maybe making its type be `FlowNodeType | <whatever types for edges>)` too, but maybe not necessary for now

src/web/topic/components/Surface/TopicToolbar.tsx
```diff
@@ -117,6 +121,22 @@ export const TopicToolbar = () => {
             >
               <Redo />
             </IconButton>
+
+            <Divider orientation="vertical" />
+
+            <IconButton
```
keyserj, line 127: can we also wrap this in `{userCanEditTopicData && [component]}` so that it doesn't show up if the user can't edit the topic?

keyserj (commented):
Looks flawless except the one check for user editability! Easy to miss, but playground topics (just saved to your browser's localstorage) and saved-to-database-and-therefore-shareable topics behave a bit differently

src/web/topic/components/Surface/TopicToolbar.tsx
seanyeh: it's already gated under https://github.com/amelioro/ameliorate/blob/d876bdcdfd46f9e262b0d82df7777b4846ca96c5/src/web/topic/components/Surface/TopicToolbar.tsx#L91 right?

keyserj: Omg you're right 😅 nice

keyserj (approved):
Awesome work!
giphy

keyserj:
@all-contributors please add @seanyeh for code

PR #281, opened 2023-12-05
Add top level claims to details

### LINKED ISSUE #271: Add top-level claims to details pane
It's nice that we have an indicator to tell if there are claims for a node/edge, but it'd be even nicer if I could see what kinds of claims are being made without going to the claim tree. Once I go to the claim tree, I leave the context of the topic diagram, and I only want to do that when I'm really focusing on a node/edge.

We could add the top-level supports and critiques to the details pane, in two columns. Something like:
image

- Should be able to add nodes from here, perhaps an "add" button in each column
- Clicking on nodes in the details pane shouldn't select them - we should be able to edit the text/score of a node from this pane without making the pane disruptively switch to that node's details
- Sort these by score if there is a score (only on render, so it's not disrupted after score selection?)

Questions:
- How to fit two columns in the pane? It's currently 300px wide and each node is 176px minimum width. We want to support 360px phones solidly, and 320px phones at least functionally. Perhaps drawer could be 352px (two nodes) + 4px padding, and have the collapse icon move into the drawer itself?
- Should the "add" button go at the top of a column or bottom? Top would never shift, but bottom would make the new node appear right below your click, which would be nice

UI thoughts:
- Probably cap the height of the columns and force scrolling after like 3-4 nodes?
- Might not be motivated to click on the claim tree - perhaps a number could display how many claims aren't in the displayed top-level, or a "view more" link that only shows if you aren't viewing all of them

**Additional context**
- This does hide the edges between the top-level claims and the root claim, but those don't seem to be used much at this point. Can re-evaluate UI if those start seeming more useful.

**Technical ideas**

keyserj:
Closes #271 

### Description of changes

- see commit messages

### Additional context

- didn't cap height of claims section because we'll probably just extract this to a separate details tab later
- didn't sort claims by score because it might be jarring for a node to jump after being scored differently. probably animation would alleviate this, but didn't want to deal with that yet
- big tech debt logic added specifically to root claims, so we should replace root claims with edges directly to argued diagram part as soon as we can


keyserj:
- details indicator does not indicate for claims that have their own claims, but fixing this is annoying until we remove root claim, so we're gonna leave it for now
- claim tree indicator does not indicate if a claim has a claim and the parent claim is deleted, leaving the child claim without a parent, but this is fine because it's a pretty unlikely situation and slightly annoying to fix

PR #338, opened 2024-02-19
feat: take screenshot of diagram

### LINKED ISSUE #332: Download image of diagram
Sometimes, after putting a lot of effort into a topic, I think it'd be cool to take a high-quality screenshot of the diagram, perhaps to print and put on a wall, or set as a desktop background. I can take a screenshot myself using something like snipping tool or ShareX, but it's tough to get those to be high quality.

Button in the `MoreActionsDrawer` that downloads a high-quality image of the currently-displayed diagram:
image
image
(if there are too many icons in this row, the icons to the right of the image icon can be moved to the next row)

- Material UI's [image icon](https://mui.com/material-ui/material-icons/?query=image&selected=Image) seems good for the icon
- Potentially we could allow the user to enter their desired dimensions, but I think hardcoding 2560 x 1440 px is good enough
- User should be able to use this button even if not logged in, similar to the existing download button

Concerns:
- Allowing a width x height to be specified and zoomed to sounds optimal, but we need access to the flow's displayed nodes, and we would need to hack a way to get that from outside of the react flow provider

Taking a screenshot myself using something like snipping tool or ShareX - but it's hard to get these to be high enough quality, especially if you're on a small monitor. Also, ideally background details would be dropped (like the little dots, collapse drawer button, etc.), and I'd have to use devtools to remove these for a manual screenshot.

**Additional context**
Diagrams are supposed to be able to evolve, so a screenshot would probably go out of date at some point, but that's ok.

**Technical ideas**
- React-flow has a perfect [example](https://reactflow.dev/examples/misc/download-image) of how to do this
  - might need to use the diagram id in the query selector, like ``querySelector(`#${activeView} .react-flow__viewport`)``, because technically the topic diagram and explore diagram [are both always displaying](https://github.com/amelioro/ameliorate/blob/39da8cd1522f73e5bae0f121f9c63aaf41af6999/src/web/topic/components/TopicWorkspace/TopicWorkspace.tsx#L52-L72)

keyserj:
Closes #332 

### Description of changes

-

### Additional context

- requested from discourse session

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,7 @@
         "@trpc/server": "10.34.0",
         "elkjs": "0.8.2",
         "framer-motion": "^10.16.5",
+        "html-to-image": "^1.11.11",
         "immer": "10.0.2",
         "js-file-download": "0.4.12",
         "lodash": "4.17.21",
@@ -18086,6 +18087,11 @@
       "funding": {
         "url": "https://github.com/sponsors/sindresorhus"
       }
+    },
+    "node_modules/html-to-image": {
+      "version": "1.11.11",
+      "resolved": "https://registry.npmjs.org/html-to-image/-/html-to-image-1.11.11.tgz",
+      "integrity": "sha512-9gux8QhvjRO/erSnDPv28noDZcPZmYE7e1vFsBLKLlRlKDSqNJYebj6Qz1TGd5lsRV+X+xYyjCKjuZdABinWjA=="
     },
     "node_modules/html-webpack-plugin": {
       "version": "5.5.0",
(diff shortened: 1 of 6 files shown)


keyserj:
I initially went the path of screenshotting the diagram as-seen, rather than centering and sizing to a width x height, but sizing by width x height seems ideal for scenarios where you want to print the diagram for something like a picture frame. (right now you can't specify width x height but will add that in the future).

Here's the code for capturing the diagram as seen, and enhancing the pixel quality:
```
const element = document.querySelector(".react-flow__viewport");
    console.log(element.clientWidth, element.clientHeight);
    const widthRatio = Math.ceil(2500.0 / element.clientWidth);
    const heightRatio = Math.ceil(2500.0 / element.clientHeight);

    toPng(element, {
      backgroundColor: "#fff",
      pixelRatio: Math.min(widthRatio, heightRatio),
    }).then(downloadImage);
  };
```

PR #394, opened 2024-04-22
Add filter to url params

### LINKED ISSUE #382: Put filter options into URL params
### Describe your issue

It's annoying when I refresh the page and lose my filter. It'd also be nice if I could send a link to someone and they could see exactly what I see - which isn't the case if they don't get my filter.

### Solution you'd like

Encode filter options into URL params

### Questions to answer

- what format to use? would be nice if the params were human readable, but ideally we wouldn't need to manually update the code to include new filter options in the URL params, so a generalized algorithm for encoding is probably best

- saving to local storage; this doesn't enable sharing filters with others via URL, or using back to go to previous filter, but could be a short-term solution to save filters beyond refresh

### Additional context

Wait until multiple filters are implemented? Since that might affect how filter options should be encoded in the URL params?

### Technical ideas

- with a generalized encoding algorithm
  - it'd probably be ideal to only add filter options that are relevant for the selected type. e.g. right now, with type = "none", there might still be a `centralProblemId` set, which is nice for remembering params when switching between filters, but is a bit much for putting in the URL
  - still would prefer to shorten uuids

### Technical questions

- how will store schema refactoring be handled when schema is put into the URL? ideally there wouldn't be much work to ensure old schemas still work

keyserj:
Closes #382 

### Description of changes

-

### Additional context

-

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,7 @@
         "html-to-image": "^1.11.11",
         "immer": "10.0.2",
         "js-file-download": "0.4.12",
+        "jsonrepair": "^3.6.1",
         "lodash": "4.17.21",
         "material-react-table": "1.14.0",
         "microdiff": "1.3.2",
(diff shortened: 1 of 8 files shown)


keyserj:
See https://github.com/amelioro/ameliorate/issues/382#issuecomment-2070928504

PR #429, opened 2024-05-27
Fix upload doesnt maintain view topic

keyserj:
### Description of changes

-

### Additional context

-


keyserj:
Note: wouldn't have missed this if we used something more sensical than quickViewStore.topic.id to determine if store should sync with api (store should probably have a config on it to setSync(off) and setSync(on) before making an action we don't want to sync, similar to how temporal store works).

PR #459, opened 2024-07-02
Add sentry monitoring

### LINKED ISSUE #293: Add system logging
As a developer, if something breaks in production and I want to investigate it, but I don't know yet how to reproduce locally, it's really hard to investigate without having logs.

Look into logging solutions. There should be something that doesn't require adding many lines throughout the codebase. Some desirables:
- logs should be reachable even if the app can't be loaded
- store for 7+ days
- log console-logged errors with messages and stack traces
- maybe with tracing? to track where an error happened between client, server, which endpoint
- getting infrastructure logs would be great too, not sure if that's feasible (e.g. netlify lambdas, which only store logs for 1 day)

**Additional context**

**Technical ideas**
Some suggestions from HackerNews: [HyperDX thread](https://news.ycombinator.com/item?id=37558357)

keyserj:
Closes #293 

### Description of changes

-

### Additional context

-

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,3 +10,8 @@ AUTH0_CLIENT_SECRET="bar"
 AUTH0_SECRET="de4825ece64ee5db0436de97c507b53da99ee2aa7946179f139b084caf755529"
 
 SENDGRID_API_KEY=""
+SENTRY_AUTH_TOKEN=""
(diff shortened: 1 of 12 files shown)


keyserj:
note: adding source maps for sentry monitoring seems to add 30s to build time

PR #506, opened 2024-09-16
chore: log request urls for netlify functions

keyserj:
### Description of changes

-

### Additional context

-


keyserj:
It seems like nextjs middleware might run on netlify edge instead of serverless functions? Not sure why this isn't working. But it's not worth spending much time on this right now, so I'm gonna close this PR.

PR #538, opened 2024-10-20
fix: changes lost on navigate back

keyserj:
to reproduce:
1. create or open a topic you can edit
2. make some edits (add/remove nodes)
3. go to settings page
4. use back button to go back to your topic
5. observe your edits have been undone

this happened because the API request to load data for a topic was being kept fresh forever, so it would never re-request without a page refresh. once loaded with old data, the old data was saved, because all state changes are saved.

in this case, navigating using the back button would perform an SPA page navigation, reusing the data it was previously loaded with.

this bug was introduced by #441

### Description of changes

-

### Additional context

-


keyserj:
FYI @prototyperspective this fixes the issue you ran into with losing your data

prototyperspective:
Yes, great to see such a quick fix!

PR #555, opened 2024-11-01
Misc touchups 9

keyserj:
### Description of changes

-

### Additional context

-


keyserj:
I think the main part I'm uncertain about is the circle + shadow -> square + border for score. I guess we can always revert anyway.

Before:
image

After:
image

Issue #568, opened 2024-11-24
Make spline edges end in vertical slopes


keyserj:
### Describe your issue

Recently, functionality was added to draw edge paths based on the output of our layout algorithm. Unfortunately, the layout output does not result in vertical slopes at the start and end points of each path. These vertical slopes often seem desirable, because they go with the top-down flow of the diagram, and without them, the edges can look a little haphazard.

These simple edge paths all end with arrows directly upwards, which is nice:
![image](https://github.com/user-attachments/assets/11066fdf-1fad-4ce0-9661-7443c4949302)

These edge paths based on the layout algorithm have some of their paths starting/ending at sharp angles, which is awkward:
![image](https://github.com/user-attachments/assets/b6e96af9-be96-4bfa-be0d-1f6f482cc019)

### Additional context

If we did this, I think we should be able to remove the "Draw simple edge paths" config, and just _always_ use the layout calc's edge paths.

PR #569, opened 2024-11-24
Avoid edge overlap

### LINKED ISSUE #559: Investigate options for reducing clutter of edges
### Describe your issue

Edges seem to add a lot of clutter when there are more than a few of them:

- edge boxes overlap often
- edge boxes have text and indicators in them

### Solution you'd like

Not entirely sure what's best. Potentially

- text and indicators could conditionally be shown e.g. via #478 
- by default only text could show
  - with the box and its background? without?
- edge label placement could be improved to avoid other labels

_No response_

### Additional context

_No response_

- the last time I investigated label placement via the [layout library that Ameliorate uses](https://github.com/kieler/elkjs), I think I had misunderstood a few of the library's options, and was creating more space than necessary. I had forgotten that right now the edge label is just [dumbly placed based on a bezier curve between the two nodes it's connecting](https://github.com/amelioro/ameliorate/blob/d5f1470c4429dbb7d06bd87a1fa2f0175f26008d/src/web/topic/components/Edge/ScoreEdge.tsx#L108-L115), without using the output of the layout algorithm which would account for other nodes/edges 😅

keyserj:
Closes #559 

Unfortunately I spent a ton of time trying to get edges and their labels to avoid overlapping, in order to reduce the feeling of clutter from edges. I couldn't get through all of the issues, and I didn't think it was worth spending more time on this right now (spent more than I should have), so I've added two separate config options to enable the functionality that might benefit some situations but not all situations:
image

The first config is "Draw simple edge paths". It defaults to true (which is the current-prod behavior), but when set to false, edge paths are drawn based on the output from the layout algorithm, so they do a better job of avoiding each other.
- The main problem with this (and the reason the config exists) is that the layout output does not result in vertical slopes at the start and end points of each path. These vertical slopes often seem desirable, because they go with the top-down flow of the diagram, and without them, the edges can look a little haphazard. Created #568 to address this.

The second config is "Avoid edge label overlap". It defaults to false (which is the current-prod behavior), but when set to true, edge labels are included in the layout algorithm, which results in them trying to avoid overlapping with each other.
- The main problem with this (and the reason the config exists) is that the ELK layout algorithm treats edge labels as nodes, which results in inconsistent spacing between node layers, as well as more node layers than otherwise. These both can result in a more chaotic layout (though sometimes it's fine). See https://github.com/eclipse/elk/issues/1092 for more info - if that is resolved, we may be able to always include edge labels in the layout, without this config.


keyserj:
One example of before:
image

vs after:
image

prototyperspective:
Great, much better now!
1. Couldn't one simply add a few pixels or a padding to the labels / label-widths?
2. I think it would be best to make this option enabled by default.
3. Nodes can still overlap labels: e.g. [here](https://ameliorate.app/Prototyperspective/CarsAreHarmfulToSociety) without the option enabled "One can stop incentivizing and facilitating car" in bottom left overlaps a label and "This requires collective investment" as well as "One could develop a system that checks travel times", "Climate change has a lots of current" and "This would save many years " overlaps edges but with the option enabled "Climate change has a lots of current and future negative effects on health" still overlaps the "subproblem of" edge (on my end at least) and "Electric vehicles only pollute during[…]" overlaps "addresses".

keyserj:
> Couldn't one simply add a few pixels or a padding to the labels / label-widths?

Yeah I'll probably just do that. I originally was thinking that an average of all the label widths made more sense, to balance between using too much space vs using too little space, but I think that's not the case. It doesn't seem like a big deal to use too much space, whereas labels still overlapping basically forfeits a lot of the value of trying to avoid overlap in the first place!

> I think it would be best to make this option enabled by default

I'll keep an eye on how it looks as I continue using more diagrams, but my impression is currently that the creation of inconsistent spacing between node layers, as well as more node layers than before, forfeits a lot of the value that the node layers create in the first place: of being able to quickly look at an organized-feeling diagram and know where you need to look to find section of node types you want to see. Here's the [cars-going-too-fast example](https://ameliorate.app/examples/detailed-cars-going-too-fast) that hopefully illustrates what I mean:

image

image

> Nodes can still overlap labels

I think in your example you need to turn off the simple edge config I added in the More Actions Drawer:
image

It is pretty annoying to have these as separate configs but I explain my reasoning for it in the PR description.

With that config set, edge labels don't overlap, as far as I can see:
image

keyserj:
> ultimately it probably needs to improved so that setting can be a default

Yeah I agree, it's not ideal when it has to be manually configured to get the benefits

> For the sharp angles, I think it would be best anyway if the lines connected to different locations of the node so that one can display the edge rating directly above/underneath the node.

That's a good point. I hadn't considered that separate connections (node "ports" is the term the layout library uses) would be desirable - if we want to do something like that edge-ratings-closer-to-nodes idea, we'd probably want the ports to be separate. One consideration though is that if there are 8+ ports on one side of a node, that would probably get super hectic (though such a case might want to be avoided/refactored by the topic author anyway).

PR #570, opened 2024-11-25
touchup: increase edge label width for layout

keyserj:
Could use the average-width to try and balance using too much space vs using too little space, but it doesn't seem like a big deal to use too much space, whereas labels still overlapping basically forfeits a lot of the value of trying to avoid overlap.

Related #569

### Description of changes

-

### Additional context

-


keyserj:
Looks better imo and extra space doesn't seem bad - here are some before/after pics:

[climate-change](https://ameliorate.app/examples/climate-change)
image
image

[cars-going-too-fast](https://ameliorate.app/examples/detailed-cars-going-too-fast)
image
image

PR #575, opened 2024-11-25
fix: header cut off

keyserj:
### Description of changes

-

### Additional context

-


keyserj:
Doesn't consistently reproduce, not quite sure why. But I think this should fix it.

PR #596, opened 2024-12-08
touchup: set role button for diagram parts

keyserj:
so they can be clicked by vimium

### Description of changes

-

### Additional context

-


keyserj:
One idea that may be better/less-confusing is if the node type label, with it's colored & distinct background, is specifically used for the click-to-select point. Then the rest of the white background of the node can be used for clicking the input text. One downside of that though is that someone might expect clicking elsewhere on the node to select it - why just the label?

Will just keep that in mind though, this seems good enough for now.

Issue #551, opened 2024-10-31
Brainstorm ideas for improving UX of the Panes, Headers, Toolbars


keyserj:
### Describe your issue

The panes, headers, and toolbars don't seem as intuitive and sleek as they could be (related #298). Let's consider what can be improved.

Known concerns:
- Quick Views are not blatant but should be
- Filters are somewhat of a waste of space for new viewers who probably won't use them
- two headers take up a lot of space from the height of the diagram
- not obvious which Topic we're on if not viewing Topic Details
- hard to tell which toolbar buttons I should care about
- textboxes in the topic details don't line up (watch, description)
- read-only textboxes don't really look readonly (probably should be "show more"/"show less")

Questions:
- how should these relate to a Simplified View like https://github.com/amelioro/ameliorate/discussions/541#discussioncomment-11052040?

Ideas:
- check out for ideas: Miro, draw.io, vscode
- could put Quick Views in a separate, simpler pane, maybe accessible via a views button in the toolbar
- maybe the toolbar should have more menu buttons, rather than putting every extra thing in one "extra things" menu
- maybe nodes in Details pane can fit in one line (not only take up less space, but distinguish from selected node)

keyserj:
Initial design idea (main inspiration: https://miro.com/ and https://www.tldraw.com/):
![image](https://github.com/user-attachments/assets/86a4dd72-43e0-45c7-b65b-67c1b2570891)

keyserj:
Made a Figma prototype for this - it's much easier to modify than a ShareX screenshot 😄, plus you can click around a little and easily see the different smaller-screen views.
- prototype: https://www.figma.com/proto/iSM88wgFYeiQA35oAPoOgF/Interface-redesign
- design (probably shouldn't need to see this outside of maintaining the prototype): https://www.figma.com/design/iSM88wgFYeiQA35oAPoOgF/Interface-redesign

I think the work involved here is much smaller than I originally anticipated. Not planning on moving the contents of the Topic Views pane nor of the More Actions Drawer; potentially putting into a text-based file menu could make things easier to find & discover, but that's another chunk of effort that seems lower priority right now. Also not going to change the Topic Details/Settings right now.

Expected work chunks:
- Create workspace overlay for headers & quick views
  - overlay Topic author/title & Quick Views Select onto workspace
  - move Site Header & Workspace Header into workspace overlay
  - split toolbar into overlayed Side Toolbar and Main Toolbar
- Make Topic Pane take up full screen for small screens
[...]

PR #607, opened 2024-12-16
Use tabs for details pane

### LINKED ISSUE #298: Reduce amount of stuff shown by default for details pane
The details pane can get pretty crowded when there are a few things in each section. This is really really rough on mobile, but also can be bad on desktop
image

Each section should probably just be put into a tab, and when a tab is selected, a param should be set for it in the URL (see #214)
image

Maybe something like vscode
image

Questions:
- Should the tabs be vertical or horizontal? It seems like a lot to double up the [details|views] tabs with this

**Additional context**

**Technical ideas**

keyserj:
Closes #298 

### Description of changes

- see commit messages

### Additional context

- topic details hasn't been changed and still feels awkward, will try to think about that as part of #551

### DIFF
--- a/src/web/comment/store/commentStore.ts
+++ b/src/web/comment/store/commentStore.ts
@@ -240,5 +240,5 @@ export const viewComment = (commentId: string) => {
   const commentParentGraphPartId =
     threadStarterComment.parentType === "topic" ? null : threadStarterComment.parentId;
   setSelected(commentParentGraphPartId);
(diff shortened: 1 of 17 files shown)


keyserj:
This makes the ontology topic not show justification/research anymore in its default view. I think that's ok, but we should have a way to share a Topic with a specific diagram part's details tab selected, e.g. to specifically show a part's Questions. Created #608 for this.

Issue #244, opened 2023-10-06
Critical path automated testing


keyserj:
**Is your feature request related to a problem? Please describe.**
It's annoying to manually test the basic features of Ameliorate for every little change. Particularly for dependency updates, now that we've automated those with Renovate.

**Describe the solution you'd like**
Add automated testing that runs on PR for confirming that the main features of Ameliorate still work. Ideas:
- home page (take screenshot)
- log in
- create topic
- create problem, 2 criteria, 3 solutions
- score 2
- delete an unscored node
- undo/redo
- create support, critique
- screenshot: claims tree
- screenshot: topic diagram
- screenshot: criteria table

**Describe alternatives you've considered**

**Additional context**

**Technical ideas**
- playwright?

keyserj:
Running through the tutorials could be good tests too, related #419

PR #623, opened 2025-01-01
Basic e2e tests

keyserj:
Partially addresses #244 

### Description of changes

-

### Additional context

-


keyserj:
Wow that was a lot of work. Created these tickets for some issues that I ran into:
- git-auto-commit-action triggers more CI but it shouldn't... created netlify [(skip ci) request](https://answers.netlify.com/t/skip-ci-in-commit-message-should-prevent-ci-that-runs-on-pr/135971)
- for debugging it'd be nice to have playwright's snapshot diffs even if we used `--update-snapshots` to update then... created playwright issue: [ability to save snapshot diffs with --update-snapshots](https://github.com/microsoft/playwright/issues/34197)
- For some reason, Netlify preview/production hosts images based on nextjs device sizes https://github.com/vercel/next.js/issues/44244#issuecomment-2305576047, BUT local images are hosted based on the actual image size... So when the browser renders the images within the size of the viewport, the images end up compressing slightly different pixels, creating a significant diff... probably should file an issue for nextjs

Issue #520, opened 2024-09-30
Hotkey for delete selected node/edge


keyserj:
### Describe your issue

It's nice that we've got a toolbar button and right-click menu option to delete a graph part (node/edge)
![image](https://github.com/user-attachments/assets/197dae0b-01fb-48d3-adfa-842aefd42f4a)

but if my hands are on the keyboard, I lose hundreds of milliseconds moving my hands to the mouse in order to click one of those options.

### Solution you'd like

If a node or edge is selected, pressing Delete should delete it.

Note: if typing in the node's textbox, delete should do the typical text-editing, not delete the node.

### Technical ideas and questions

Should be very similar to [setting the score for the selected node/edge](https://github.com/amelioro/ameliorate/blob/a9080cee51cb396313653cb4e007388bb62d5bc3/src/web/topic/components/TopicWorkspace/TopicWorkspace.tsx#L29-L39), except:
- instead of doing something with the score, we need to [delete the graph part](https://github.com/amelioro/ameliorate/blob/a9080cee51cb396313653cb4e007388bb62d5bc3/src/web/topic/store/createDeleteActions.ts#L370-L376)
- we should also only continue with deletion if the user has permission to; should be able to check via the [canEdit hook](https://github.com/amelioro/ameliorate/blob/a9080cee51cb396313653cb4e007388bb62d5bc3/src/web/topic/components/Node/EditableNode.tsx#L33-L34)

keyserj:
To avoid duplicate efforts, I'm just going to note here that @sarccasm is currently working on this via #637.

lovelypangotra2003:
Can I work on this?

keyserj:
Hey @lovelypangotra2003 yep!

Just an FYI - I did intend https://github.com/amelioro/ameliorate/labels/good%20first%20issue s as an approachable way for new contributors to make a first contribution, but for contributors to move towards https://github.com/amelioro/ameliorate/labels/good%20second%20issue s as more-challenging tasks after their first few contributions. I realize that wasn't mentioned in CONTRIBUTING.md, so I updated that.

lovelypangotra2003:
I see, I will make a transition after looking at the current issues. I would also like to work on the backend issues. Is there a tag for that that I should look for?
Can you also review my PR, please?

keyserj:
[...]

PR #637, opened 2025-01-11
Added hotkey for deleting nodes

sarccasm:
Closes #[520]

### Description of changes
Added a keyboard shortcut (Delete) for deleting nodes in the graph editor.
Implemented useEffect with an event listener for keydown in DeleteNodeMenuItem.tsx.
-

### Additional context
Verified that nodes can be deleted using both the context menu and the Delete key.
-

### DIFF
--- a/src/web/common/components/ContextMenu/DeleteNodeMenuItem.tsx
+++ b/src/web/common/components/ContextMenu/DeleteNodeMenuItem.tsx
@@ -1,3 +1,5 @@
+import { useEffect } from "react";
+
 import { ContextMenuItem } from "@/web/common/components/ContextMenu/CloseOnClickMenuItem";
 import { useSessionUser } from "@/web/common/hooks";
 import { deleteNode } from "@/web/topic/store/createDeleteActions";
@@ -7,6 +9,20 @@ import { Node } from "@/web/topic/utils/graph";
 export const DeleteNodeMenuItem = ({ node }: { node: Node }) => {
   const { sessionUser } = useSessionUser();
   const userCanEditTopicData = useUserCanEditTopicData(sessionUser?.username);
+
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (event.key === "Delete" && userCanEditTopicData) {
+        deleteNode(node.id);
+      }
+    };
+
+    document.addEventListener("keydown", handleKeyDown);
+
+    return () => {
+      document.removeEventListener("keydown", handleKeyDown);
+    };
+  }, [node, userCanEditTopicData]);
 
   if (!userCanEditTopicData) return <></>;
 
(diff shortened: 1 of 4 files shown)


src/web/common/components/ContextMenu/DeleteNodeMenuItem.tsx
```diff
@@ -8,6 +10,20 @@ export const DeleteNodeMenuItem = ({ node }: { node: Node }) => {
   const { sessionUser } = useSessionUser();
   const userCanEditTopicData = useUserCanEditTopicData(sessionUser?.username);
 
+  useEffect(() => {
```
keyserj, line 13: This is a good idea, but since this `useEffect` is defined within the `DeleteNodeMenuItem`, the `delete` key only works when this menu item is rendered, i.e. when right-clicking on a node.

Check out the Technical Details section of the issue https://github.com/amelioro/ameliorate/issues/520 - the code referenced there shows an established pattern for adding hotkeys in Ameliorate. There, we add a hotkey hook to the `TopicWorkspace`, which is a parent container of the `Diagram`, so the hotkey will always be usable without needing any other specific components rendered.

keyserj (commented):
Welcome to Ameliorate 🙂

src/web/topic/components/TopicWorkspace/TopicWorkspace.tsx
```diff
@@ -26,6 +27,16 @@ import { getSelectedGraphPart, setSelected, useFormat } from "@/web/view/current
...
 
+const getSelectedNodeId = (): string | null => {
+  const selectedPart = getSelectedGraphPart();
+  return selectedPart ? selectedPart.id : null;
+};
+
+const nodeExists = (nodeId: string): boolean => {
+  const selectedPart = getSelectedGraphPart();
+  return selectedPart ? selectedPart.id === nodeId : false;
+};
```
keyserj, line 38: I don't think these methods are needed. We can just `return` from our hook early if the `getSelectedGraphPart()` is null, like how the hotkey for scoring behaves:

https://github.com/amelioro/ameliorate/blob/778fcf2f039ec1f2686eb70dac0c6e773f99f9bf/src/web/topic/components/TopicWorkspace/TopicWorkspace.tsx#L35-L36

And if the selected part isn't null, we can pass it directly to the `deleteGraphPart` method.

src/web/common/components/ContextMenu/DeleteNodeMenuItem.tsx
keyserj: The code you've added to `TopicWorkspace` is looking better, but this `DeleteNodeMenuItem` code still needs to be undone.

keyserj:
There haven't been updates on this in a while so I'm going to close it - let me know if you're still working on it and we can reopen!

PR #638, opened 2025-01-11
Fix headline size

keyserj:
### Description of changes

-

### Additional context

-


keyserj:
I'm very unsure why the previously-committed screenshot for the playground has text (e.g. the topic title "Playground Topic") that changed in spacing between the characters ("font kerning"?).

Run that committed the previous screenshot: https://github.com/amelioro/ameliorate/actions/runs/12718000506/job/35455606217

Run that committed the new screenshot in this PR: https://github.com/amelioro/ameliorate/actions/runs/12728477610/job/35479098295?pr=638

It looks like the Docker version, container's OS version (and image), browser version (from request headers recorded via playwright-report traces), npm/node versions, are all identical.

From the traces, I noticed that the previous run had one woff2 font request, and this PR's run had four woff2 font requests. I thought maybe this was due to a Netlify change, but the next-netlify plugin is also the same version between runs. I'm not sure if these font requests could be the cause of the diff.

This Playwright issue has multiple people who encountered font kerning issues with WebKit https://github.com/microsoft/playwright/issues/20203, but we're using chromium. Perhaps one of the solutions mentioned in that issue can be considered for Ameliorate.

I prefer not to spend more time on this right now - I'll probably look into it more when it happens again.

PR #643, opened 2025-01-17
touchup(change node type drop down menu): applied icons and colors

### LINKED ISSUE #636: Add colored icons to "change node type" menu
### Describe your issue

When I want to change a node from one type to another, I use the context menu (by right-clicking the diagram background), but it takes a few seconds to find the type in the list that I want to change to. The list is kind of sorted by information category (breakdown, justification, research), but this isn't obvious, nor is it helpful enough. Please make it easier to quickly pick the node type I want from the context menu.

Current menu:

image

### Solution you'd like

Same solution as the one that ends up being used in #629 (for consistency)

_No response_

### Additional context

_No response_

_No response_

rdubbi:
Closes #636 

### Description of changes

- closely mirrored code developed during #633 for AddNodeMenuItem.tsx and applied it to ChangeNodeTypeMenuItem.tsx.

### DIFF
--- a/src/web/common/components/ContextMenu/ChangeNodeTypeMenuItem.tsx
+++ b/src/web/common/components/ContextMenu/ChangeNodeTypeMenuItem.tsx
@@ -1,3 +1,4 @@
+import { useTheme } from "@mui/material/styles";
 import { NestedMenuItem } from "mui-nested-menu";
 
 import { getSameCategoryNodeTypes } from "@/common/node";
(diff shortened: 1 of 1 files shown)


src/web/common/components/ContextMenu/ChangeNodeTypeMenuItem.tsx
```diff
@@ -17,27 +18,29 @@ export const ChangeNodeTypeMenuItem = ({ node, parentMenuOpen }: Props) => {
...
+  const theme = useTheme();
+
   if (!userCanEditTopicData) return <></>;
 
   return (
-    <>
-      <NestedMenuItem
-        label="Change node type"
-        parentMenuOpen={parentMenuOpen}
-        // match default mui menu padding and size
```
keyserj, line 27: This all looks solid to me, except it looks like you've removed the comment here `// match default mui menu padding and size`. Did you intentionally do that? The comment seems to me that it still applies, as it explains the reasoning for the specific `className`s used on the next line.

rdubbi: Sorry about that! I must have thought that I put it there and removed it. I will add it back in.

rdubbi:
I added the comment back as discussed but got a commit lint error:

`Error: You have commit messages with errors

⧗   input: Update ChangeNodeTypeMenuItem.tsx

Added original comment back per request.
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]`

It is not clear to me if making the changes to resolve the error is necessary, and if it is, what the best way to make the changes are. 
I could not seem to find a good way to edit the commit message on github. All I could find was a bad way: adding a small change to the file (like punctuation at the end of the comment I added back in) that would allow me to modify the commit message again. I also suspect making the change locally and pushing up the code again is not ideal. 

Please let me know if the commit message amendment is necessary to accept the updated code, and if it is, the best way to change it without making a mess! Thank you!

keyserj (approved):
Thanks for this PR! No need to fix the commitlint error - when I squash the commits, I'll use the first commit you've got and that one fits the conventional commits format just fine.

> I could not seem to find a good way to edit the commit message

> I also suspect making the change locally and pushing up the code again is not ideal

As an FYI: I don't think the GitHub UI allows editing commit messages, but making the change locally and pushing that would be fine. Generally if work can be done outside of your local, that's when you want to avoid modifying existing commits and force-pushing, because that replaces the remote with your local and can remove other people's changes. But if it's just you, and you make sure to pull any changes you make through GitHub's UI, then it's no issue. You can check out GitHub's [Changing a commit message](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message) documentation for how to change the message locally.

PR #646, opened 2025-01-21
touchup: improve node label-icon alignment

keyserj:
### Description of changes

- see commit messages

### Additional context

first commit attempts to better center-align the node label and node icon (note that FF already looked good and still seems to):

image

second commit attempts to make spacing more consistent around the node label and icon:

image

third commit removes the very thin shadow that sometimes would render on the bottom/right of node label:

image


keyserj:
before/after (Chrome):

image
image

before/after (FF):

image
image

PR #647, opened 2025-01-23
Touchup homepage

keyserj:
### Description of changes

- see commit messages

### Additional context

-


keyserj:
before/after re: table header changes
image
image

Issue #482, opened 2024-08-13
"Reset topic" should delete comments too


keyserj:
### Describe your issue

If I'm trying to start over with a new diagram by clicking Reset, existing comments probably won't be relevant anymore, and it's really annoying when Reset doesn't delete these comments, because then I have to do so manually.

Right now it looks like this, keeping comments around after reset:

https://github.com/user-attachments/assets/86a242a9-7886-4ac0-9de3-915557badb8d

### Solution you'd like

When the reset button is clicked, remove comments along with nodes and edges.

But since comment deletion cannot be undone, let's also conditionally show a dialog if the topic has comments that'll be deleted, saying something like "This topic has comments that will be deleted by this action, and undo will not restore these comments. Continue resetting?" Should be very similar to deleting a topic via the topic's settings page:
![image](https://github.com/user-attachments/assets/e6275b23-5bd2-4fe2-896b-517368ef4cd0)

### Additional context

Originally comments hadn't been deleted with reset because they aren't undo-able (whereas deleting nodes/edges can be undone), but I think we'd still be better off deleting the comments. Showing a dialog seems like a good enough warning.

### Technical ideas and questions

Should be very similar to the code for deleting a topic [here](https://github.com/amelioro/ameliorate/blob/b08ee70bc658c9a73d636fa05a8d285a0bd7a0ee/src/web/topic/components/TopicForm/TopicForm.tsx#L137-L174), except instead of making an api request on delete, we'll just be calling a `reset` method from the `commentStore` and `draftStore`.

2div:
Hi @keyserj ,

I want to try to work in this issue.
I am not familiar to this app and I would appreciate any help you can give me to solve this issue.
Can you advise me where is the file of the reset topic button and what are the proper steps to solve this issue.

Thank you

keyserj:
Hey @2div , welcome! I'm happy to help 🙂 if at any time you feel there might be a lot of back-and-forth about a question(s) you have, feel free to message me on the [Discord server](https://discord.gg/3KhdyJkTWT) as well.

[...]

PR #664, opened 2025-03-13
feat: delete comments when reset topic

2div:
Closes #issue-number-here

Description of changes

    feat: 

    https://github.com/amelioro/ameliorate/issues/482 when Reset topic clicked it should delete comments too

Additional context

    A dded a function which will delete comments when Reset topic button clicked

https://github.com/user-attachments/assets/5987b768-670c-4a28-b5ad-3f0b93cd7ffa


src/web/comment/store/commentStore.ts
```diff
@@ -159,6 +159,10 @@ export const deleteComment = (commentId: string) => {
   );
 };
 
+export const resetComment = () => {
+  useCommentStore.setState({ ...initialState }, false, "resetComment");
```
keyserj, line 163: This comment from the previous PR https://github.com/amelioro/ameliorate/pull/580#discussion_r1865020313 still applies here

src/web/comment/store/commentStore.ts
```diff
@@ -161,6 +161,11 @@ export const deleteComment = (commentId: string) => {
   );
 };
 
+export const resetComment = () => {
```
keyserj, line 164:
```suggestion
export const resetComments = () => {
```

I think plural "comments" makes more sense for the name, because we're not just resetting one comment

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -141,6 +149,44 @@ export const MoreActionsDrawer = ({
...
+
+  const resetComments = (
+    <Dialog
+      open={resetDialogOpen}
+      onClose={() => setResetDialogOpen(false)}
+      aria-labelledby="alert-dialog-title"
+    >
+      <DialogTitle id="alert-dialog-title">
+        {/* {/ Delete topic {user.username}/{topic.title}? /} */}
+        Reset Comments
```
keyserj, line 162: It's good that we're warning about comments in the dialog's text because they can't be undone, but the action still will be resetting everything (i.e. also topic data + quick views) so I think the title makes sense more as "Reset Topic".

keyserj: (also the commented-out "Delete topic" line should be removed)

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -141,6 +149,44 @@ export const MoreActionsDrawer = ({
...
+          color="error"
+          variant="contained"
+          onClick={() => {
+            setResetDialogOpen(false);
+            resetComment();
+            resetTopicData();
+            resetQuickViews();
+          }}
+        >
+          RESET COMMENT
```
keyserj, line 184: I think this should also be "Reset Topic" for the reason above

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -141,6 +149,44 @@ export const MoreActionsDrawer = ({
   const fillNodesWithColor = useFillNodesWithColor();
   const expandDetailsTabs = useExpandDetailsTabs();
 
+  const [resetDialogOpen, setResetDialogOpen] = useState(false);
+
+  const resetComments = (
```
keyserj, line 154:
```suggestion
  const resetDialog = (
```

This variable is the dialog component - `resetComments` is a little misleading, as if it's a method rather than dialog component

keyserj (commented):
It's looking good!! Just a few smaller things

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -194,8 +240,7 @@ export const MoreActionsDrawer = ({
                 title="Reset Topic"
                 aria-label="Reset Topic"
                 onClick={() => {
-                  resetTopicData();
-                  resetQuickViews();
+                  setResetDialogOpen(true);
```
keyserj, line 243: Oh sorry one more thing I realized - this is fine as is for now, but ideally it would check if there are actually comments to delete, and _not_ show the dialog if there are none to delete. This can be done separately from this PR or it can be done in this PR, but it does add a bit more work. Probably would need to add a `hasComments` method to the commentStore that can be used here

src/web/comment/store/commentStore.ts
2div: I agree, using the plural form makes more sense.

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
2div: Changed "Reset Comment" to "Reset Topic"

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
2div: commented-out "Delete topic" line removed

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
2div: The component name "resetComments" has been changed to "resetDialog"

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
2div: You're right, it would be better to check if there are comments before showing the dialog.I've added a hasComments function that will return true if there are more than 0 comments.

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -194,8 +237,7 @@ export const MoreActionsDrawer = ({
                 title="Reset Topic"
                 aria-label="Reset Topic"
                 onClick={() => {
-                  resetTopicData();
-                  resetQuickViews();
+                  hasComments() && setResetDialogOpen(true);
```
keyserj, line 240: Generally `if` is clearer for conditionally running logic than `&&`, because `&&` is originally intended as boolean evaluation, with execution being more of a side effect. So I would try to avoid `&&` for conditionally running something. `&&` is ok for conditionally running something specifically in the context of jsx conditional rendering, because you can only use expressions in that context, not statements like `if`. [Here's a great article on statements vs expressions](https://www.joshwcomeau.com/javascript/statements-vs-expressions/).

Also `if` is particularly good here because we should still be running the reset logic (at least for resetting the topic data and quick views) if there are no comments! We just don't need to use the dialog to warn users.

src/web/comment/store/commentStore.ts
```diff
@@ -161,6 +161,17 @@ export const deleteComment = (commentId: string) => {
   );
 };
 
+export const hasComments = () => {
+  const comments = useCommentStore.getState().comments;
+  if (comments.length > 0) return true;
+  else return false;
```
keyserj, line 167: (nitpick) in most cases like this one, I think conditionally true vs false is just extra verbose without adding readability, compared to just returning the expression that's being evaluated :

```suggestion
  return comments.length > 0;
```

keyserj (commented):
It's looking good!

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
2div: Thanks for the suggestion! That makes sense, using if statement is clearer for conditionally running logic, and it's good to ensure the topic data and quick views are reset even if there are no comments.
 I also really liked the article you shared. Based on your feedback, I've updated the code to use if statement as follows:

if (hasComments()) {
  setResetDialogOpen(true);
} else {
  resetTopicData();
  resetQuickViews();
}

src/web/comment/store/commentStore.ts
2div: "Good point! I've simplified the function as suggested. Thanks for the feedback!" 😊

2div (commented):
Refactor: simplify hasComments function and improve reset logic

keyserj (approved):
This looks ready, great work!! Thanks for your contribution 🙂

image

keyserj:
@all-contributors please add @2div for code

CatGuardian:
🎉🔥

Way to go 2div! This is absolutely incredible!

I am extremely grateful you took the time to help with this! You are awesome.

Let's go!

PR #686, opened 2025-04-10
Fix circular dependency infinite recursion

keyserj:
### Description of changes

- see commit messages

### Additional context

-


keyserj:
> Suspect Issues

This isn't a problem, I was running an old version of the code locally and this triggered.

PR #687, opened 2025-04-12
feat(screenshot): add custom resolution modal for screenshot export

### LINKED ISSUE #339: Allow user to specify resolution when downloading a screenshot
I'm happy that I can take a high quality image of my diagram via the screenshot button, but I want to be able to increase or decrease quality and change the dimensions as I want, so that I can get exactly the image I want, perhaps for a desktop background, or a print-out diagram.

When I click the screenshot button, I want to specify the resolution of the image:
image

**Additional context**
The screenshot button is in the `MoreActionsDrawer`, accessed here:
image

**Technical ideas**
- MUI's [dialog](https://mui.com/material-ui/react-dialog/#form-dialogs) and text field (with [input adornments](https://mui.com/material-ui/react-text-field/#input-adornments)) components seem like they should work well for this
- Zod + react-hook-form could also be good for ensuring the inputs are valid when submitting the dialog, similar to how we validate topic description [here](https://github.com/amelioro/ameliorate/blob/df64bf080351784b35d01085b7ba3f86e3b28476/src/web/topic/components/TopicPane/TopicDetails.tsx#L38-L50):
image

lovelypangotra2003:
Closes #339

### Allow User to specify resolution when downloading a screenshot
**Problem**

The Screenshot button in the MoreActionsDrawer was downloading the screenshot at a fixed resolution, without giving users the option to specify custom width and height. This limited flexibility for users who needed screenshots in specific dimensions.

**Solution**

- Implemented a form-based dialog that opens when the Screenshot button is clicked.
- Added input fields for width and height, allowing users to define custom dimensions.
- Used react-hook-form with Zod validation to ensure valid input.
- Automatically closes the dialog upon successful submission.

**Users can now:**

- Open a dialog on clicking the Screenshot button
- Specify custom screenshot dimensions
- Download PNGs tailored to their needs
- Get instant feedback if invalid values are entered

https://github.com/user-attachments/assets/ed338b42-0154-4dbe-8458-dd427afd5916
@keyserj  Please review my changes when you have time. Thanks!

### DIFF
--- a/src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
+++ b/src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
@@ -1,3 +1,4 @@
+import { zodResolver } from "@hookform/resolvers/zod";
 import {
   AutoAwesomeMotion,
   AutoStoriesOutlined,
@@ -29,19 +30,23 @@ import {
   Divider,
   Drawer,
   IconButton,
+  InputAdornment,
   List,
   ListItem,
   ListItemIcon,
   ListItemText,
   MenuItem,
   Select,
+  TextField,
   ToggleButton,
   Tooltip,
   Typography,
 } from "@mui/material";
 import { toPng } from "html-to-image";
 import { useState } from "react";
+import { useForm } from "react-hook-form";
 import { getRectOfNodes, getTransformForBounds } from "reactflow";
+import { z } from "zod";
 
 import { hasComments, resetComments } from "@/web/comment/store/commentStore";
 import { getDisplayNodes } from "@/web/topic/components/Diagram/externalFlowStore";
@@ -83,39 +88,14 @@ import {
   useFillNodesWithColor,
 } from "@/web/view/userConfigStore";
 
-const imageWidth = 2560;
-const imageHeight = 1440;
+const ScreenshotSchema = z.object({
+  width: z.number().int().positive({ message: "Must be a positive number" }),
+  height: z.number().int().positive({ message: "Must be a positive number" }),
+});
 
-const downloadScreenshot = () => {
-  const nodes = getDisplayNodes();
+type ScreenshotForm = z.infer<typeof ScreenshotSchema>;
 
-  // thanks react flow example https://reactflow.dev/examples/misc/download-image
-  const nodesBounds = getRectOfNodes(nodes);
-  const transform = getTransformForBounds(nodesBounds, imageWidth, imageHeight, 0.125, 2);
-  const viewportElement = document.querySelector(".react-flow__viewport");
-  if (!viewportElement) throw new Error("Couldn't find viewport element to take screenshot of");
-
-  toPng(viewportElement as HTMLElement, {
-    backgroundColor: "#fff",
-    width: imageWidth,
-    height: imageHeight,
-    style: {
-      width: imageWidth.toString(),
-      height: imageHeight.toString(),
-      transform: `translate(${transform[0]}px, ${transform[1]}px) scale(${transform[2]})`,
-    },
-  })
-    .then((dataUrl) => {
-      const a = document.createElement("a");
-
-      a.setAttribute("download", "topic.png");
-      a.setAttribute("href", dataUrl);
-      a.click();
-    })
-    .catch((error: unknown) => {
-      throw error;
-    });
-};
+const defaultResolution = { width: 2560, height: 1440 };
 
 interface Props {
   isMoreActionsDrawerOpen: boolean;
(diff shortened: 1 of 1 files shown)


src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -83,39 +88,14 @@ import {
...
-
-      a.setAttribute("download", "topic.png");
-      a.setAttribute("href", dataUrl);
-      a.click();
-    })
-    .catch((error: unknown) => {
-      throw error;
-    });
-};
+const ScreenshotSchema = z.object({
```
keyserj, line 91:
```suggestion
const ScreenshotFormSchema = z.object({
```
np: to match the naming pattern I've used for others like this

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -150,6 +130,46 @@ export const MoreActionsDrawer = ({
...
+  const {
+    register,
+    handleSubmit,
+    formState: { errors },
+  } = useForm<ScreenshotForm>({
+    defaultValues: defaultResolution,
+    resolver: zodResolver(ScreenshotSchema),
+  });
+
+  const onScreenshotSubmit = ({ width, height }: ScreenshotForm) => {
```
keyserj, line 144: I think the reason you moved this method into the component may be that it needs to call `setDialog(false)` afterwards? Since most of the method doesn't need access to the component's internals, I think it's nicer to keep it outside of the component, and just do something like the following when submitting the form:
```ts
onSubmit={() => {
  void handleSubmit(downloadScreenshot)();
  setDialog(false);
}
```

What do you think?

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -150,6 +130,46 @@ export const MoreActionsDrawer = ({
...
+  const [screenshotDialogOpen, setScreenshotDialogOpen] = useState(false);
+
+  const {
+    register,
+    handleSubmit,
+    formState: { errors },
+  } = useForm<ScreenshotForm>({
+    defaultValues: defaultResolution,
+    resolver: zodResolver(ScreenshotSchema),
+  });
```
keyserj, line 142:
```suggestion
  } = useForm<ScreenshotForm>({
    mode: "onBlur",
    reValidateMode: "onBlur",
    defaultValues: defaultResolution,
    resolver: zodResolver(ScreenshotSchema),
  });
```
I think validation on blur feels a little nicer (the default is to wait until first submit, then validate on on change after that)

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -83,39 +88,14 @@ import {
   useFillNodesWithColor,
 } from "@/web/view/userConfigStore";
 
-const imageWidth = 2560;
-const imageHeight = 1440;
-
-const downloadScreenshot = () => {
-  const nodes = getDisplayNodes();
-
-  // thanks react flow example https://reactflow.dev/examples/misc/download-image
```
keyserj, line 92: could you keep this comment in? I appreciate crediting it and being able to refer back if something doesn't work

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -479,6 +498,46 @@ export const MoreActionsDrawer = ({
           </ToggleButton>
         </ListItem>
       </List>
+
+      {/* Screenshot Resolution Dialog  */}
+      <Dialog open={screenshotDialogOpen} onClose={() => setScreenshotDialogOpen(false)}>
```
keyserj, line 503: Since we're using the Dialog to validate/submit data, I think it'd be more [semantically](https://www.w3schools.com/html/html5_semantic_elements.asp) accurate to use the `<form/>` element. Here's an example from elsewhere of how to do this with MUI's `Dialog`:

https://github.com/amelioro/ameliorate/blob/e78776951ac48d67214933b51c30bb973a02dba4/src/web/topic/components/TopicPane/QuickViewForm.tsx#L52-L63

It's slightly awkward to do, but I think worth it. Also note that:
1. onSubmit's `event` is passed to `handleSubmit` to prevent the page from reloading after submit
2. you don't need an `onClick` for the submit button because `type="submit"` makes the button click trigger the form's `onSubmit` by default

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -479,6 +498,46 @@ export const MoreActionsDrawer = ({
...
+        <DialogContent>
+          <TextField
+            fullWidth
+            label="Width"
+            type="number"
+            margin="dense"
+            InputProps={{ endAdornment: <InputAdornment position="end">px</InputAdornment> }}
+            error={!!errors.width}
+            helperText={errors.width?.message}
+            {...register("width", { valueAsNumber: true, min: 1 })}
```
keyserj, line 514:
```suggestion
            {...register("width", { valueAsNumber: true })}
```
Our zod schema's `positive()` handles this `min: 1` validation so it's not necessary here

src/web/topic/components/TopicWorkspace/MoreActionsDrawer.tsx
```diff
@@ -479,6 +498,46 @@ export const MoreActionsDrawer = ({
           </ToggleButton>
         </ListItem>
       </List>
+
+      {/* Screenshot Resolution Dialog  */}
+      <Dialog open={screenshotDialogOpen} onClose={() => setScreenshotDialogOpen(false)}>
+        <DialogTitle>Set Screenshot Resolution</DialogTitle>
+        <DialogContent>
+          <TextField
+            fullWidth
```
keyserj, line 507: np: `fullWidth` makes our inputs much larger than they need to be. We could remove it and add `className="flex flex-col"` to the `DialogContent` to keep the inputs smaller while also having them on separate lines.

keyserj (commented):
Hey welcome to Ameliorate! 🙂

This is looking really good, just have a few suggestions.

lovelypangotra2003:
Hi @keyserj! Recent updates have been pushed. Would love to hear any feedback you might have.

keyserj (approved):
Great work @lovelypangotra2003 🚀 this is perfect. Thanks for the contribution!

Now users don't have to deal with this! :

image

keyserj:
@all-contributors please add @lovelypangotra2003 for code

PR #696, opened 2025-04-26
feat: persist transpose state for CriteriaTable

### LINKED ISSUE #503: Transposing the table should be persisted after page refresh
### Describe your issue

Transposing the table can be nice when there are a lot of solutions and not a lot of criteria, because it removes the need to scroll to see the whole table:
chrome_2024-09-15_14-09-53

But if I refresh the page, the table does not remember that it was transposed.

### Solution you'd like

Persist the transpose state. I should be able to refresh the page and still have the table transposed.

_No response_

### Additional context

_No response_

- currently [the transpose state is component state](https://github.com/amelioro/ameliorate/blob/5c6cb91016fbf7ad5f1eb689abfcba3704b8e1aa/src/web/topic/components/CriteriaTable/CriteriaTable.tsx#L186) in the Criteria Table; we should be able to move this to the currentViewStore so that it gets persisted
  - the currentViewStore persists view data, and conveniently its state is used to store Quick View state too, so this should also enable saving transpose state with Quick Views
  - adding this to the store should look very similar to [how format is in the store](https://github.com/amelioro/ameliorate/blob/5c6cb91016fbf7ad5f1eb689abfcba3704b8e1aa/src/web/view/currentViewStore/store.ts#L23): [exposing a hook](https://github.com/amelioro/ameliorate/blob/5c6cb91016fbf7ad5f1eb689abfcba3704b8e1aa/src/web/view/currentViewStore/store.ts#L120-L122) for the table to render based on the current value, and [exposing a setter](https://github.com/amelioro/ameliorate/blob/5c6cb91016fbf7ad5f1eb689abfcba3704b8e1aa/src/web/view/currentViewStore/store.ts#L142-L144) so that the table can set the value

lovelypangotra2003:
Closes #503

This PR addresses the issue of the transpose state (`useSolutionsForColumns`) in the CriteriaTable not persisting across page refreshes. The following changes have been made:

**Added `useSolutionsForColumns` to `ViewState`**

- Updated the `ViewState` interface to include`useSolutionsForColumns`
- Set the default value for useSolutionsForColumns in `initialViewState`.

**Created a Hook and Setter:**

- Added `useUseSolutionsForColumns` to retrieve the current state of `useSolutionsForColumns`.
- Added `setUseSolutionsForColumns` to update the state.

**Integrated with CriteriaTable:**

- Updated the CriteriaTable component to use the `useUseSolutionsForColumns` hook and `setUseSolutionsForColumns` setter.
- Ensured the transpose state persists across page refreshes and integrates with Quick Views.

### DIFF
--- a/src/web/topic/components/CriteriaTable/CriteriaTable.tsx
+++ b/src/web/topic/components/CriteriaTable/CriteriaTable.tsx
@@ -29,10 +29,7 @@ import { useUserCanEditTopicData } from "@/web/topic/store/userHooks";
 import { getConnectingEdge } from "@/web/topic/utils/edge";
 import { Edge, Node } from "@/web/topic/utils/graph";
 import { useGeneralFilter, useTableFilter } from "@/web/view/currentViewStore/filter";
-import {
-  setUseSolutionsForColumns,
-  useUseSolutionsForColumns,
-} from "@/web/view/currentViewStore/store";
+import { setTransposed, useTransposed } from "@/web/view/currentViewStore/store";
 import { applyScoreFilter } from "@/web/view/utils/generalFilter";
 import { getSelectedTradeoffNodes } from "@/web/view/utils/infoFilter";
 
@@ -187,7 +184,7 @@ const buildTableDefs = (
  *   b. subsequent rows turn into row objects (`rowData`) that know their row header details (for filtering)
  */
 export const CriteriaTable = () => {
-  const useSolutionsForColumns = useUseSolutionsForColumns();
+  const transposed = useTransposed();
 
   const { sessionUser } = useSessionUser();
   const userCanEditTopicData = useUserCanEditTopicData(sessionUser?.username);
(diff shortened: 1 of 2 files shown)


src/web/topic/components/CriteriaTable/CriteriaTable.tsx
```diff
@@ -183,7 +187,7 @@ const buildTableDefs = (
  *   b. subsequent rows turn into row objects (`rowData`) that know their row header details (for filtering)
  */
 export const CriteriaTable = () => {
-  const [useSolutionsForColumns, setUseSolutionsForColumns] = useState<boolean>(true);
+  const useSolutionsForColumns = useUseSolutionsForColumns();
```
keyserj, line 190: This name is a bit awkward. I think we could actually use this simpler name because the table might include other node types besides solutions in the future:
```suggestion
  const transposed = useTransposed(); // no transpose means solutions are put into the columns
```

src/web/view/currentViewStore/store.ts
```diff
@@ -57,6 +57,7 @@ export interface ViewState {
    * unconditionally include edge labels in layout: https://github.com/eclipse/elk/issues/1092
    */
   avoidEdgeLabelOverlap: boolean;
+  useSolutionsForColumns: boolean;
```
keyserj, line 60: Can we move this to the very bottom (below `layoutThoroughness`) and add a comment above it `// table config` just to be clear that this config is specific to the table. Most of the other config is for the diagram, so it'll be nice to distinguish this.

keyserj (commented):
Nice this is looking good! Just a couple suggestions.

keyserj (approved):
Woot great work 🚀. Thanks for contributing!

PR #710, opened 2025-05-03
Topic title space

### LINKED ISSUE #536: When entering a map title it should accept spaces and replace them with hyphens in the URL
### Describe your issue

When creating a new topic most users will probably add a sentence with spaces. It will show an error and many will not understand what the problem is. Moreover, it shouldn't be a problem at all.

### Solution you'd like

The title with spaces could be displayed somewhere at the map and the URL get automatically changed so it replaces spaces with hyphens -

_No response_

### Additional context

Created a topic named like https://ameliorate.app/Prototyperspective/CarsAreHarmfulToSociety because I didn't think of using spaces because I didn't know this title was used for the URL. It would be better to have all topics with spaces like …/problems-of-cars

_No response_

lovelypangotra2003:
Closes #536 

This PR ensures consistent handling of spaces and dashes in topic titles across the backend and frontend. The following changes have been implemented:

- Backend Stores User-Entered Titles: The backend now stores the exact title entered by the user without converting spaces to dashes.

- Query Normalization: All queries normalize the title to dashes using the `normalizeTitle` function to ensure consistent retrieval of topics, regardless of space/dash variations.

- URL Safety: The `getLinkToTopic` function normalizes titles to dashes when constructing URLs, ensuring URL safety and consistency.

- Validation: Updated the `topicSchema` to validate titles while allowing spaces and dashes.

These changes ensure that:

- Users see the exact title they entered.

- Topics are uniquely identified and consistently retrieved.

- URLs remain safe and predictable.

### DIFF
--- a/src/api/routers/topic.ts
+++ b/src/api/routers/topic.ts
@@ -24,13 +24,21 @@ export const topicRouter = router({
       const isCreator = opts.input.username === opts.ctx.user?.username;
       const normalizedTitle = normalizeTitle(decodeURIComponent(opts.input.title));
 
-      return await xprisma.topic.findFirst({
-        where: {
-          title: { equals: normalizedTitle, mode: "insensitive" },
-          creatorName: opts.input.username,
-          visibility: isCreator ? undefined : { not: "private" },
-        },
-      });
+      const topics = await xprisma.$queryRaw<
+        { id: number; title: string; creatorName: string; visibility: string }[]
+      >`
+        SELECT id, title, "creatorName", visibility
+        FROM "topics"
+        WHERE "creatorName" = ${opts.input.username}
+        AND (
+          REPLACE(LOWER(title), ' ', '-') = ${normalizedTitle}
+          OR title = ${normalizedTitle}
+        )
+        AND (${isCreator} OR visibility != 'private')
+        LIMIT 1
+      `;
+
+      return topics[0] ?? null;
     }),
 
   /**
@@ -48,12 +56,34 @@ export const topicRouter = router({
     .query(async (opts) => {
       const isCreator = opts.input.username === opts.ctx.user?.username;
       const normalizedTitle = normalizeTitle(decodeURIComponent(opts.input.title));
+      const topics = await xprisma.$queryRaw<
+        {
+          id: number;
+          title: string;
+          creatorName: string;
+          description: string;
+          visibility: string;
+          allowAnyoneToEdit: boolean;
+          createdAt: Date;
+          updatedAt: Date;
+        }[]
+      >`
+        SELECT id, title, "creatorName", description, visibility, "allowAnyoneToEdit", "createdAt", "updatedAt"
+        FROM "topics"
+        WHERE "creatorName" = ${opts.input.username}
+        AND (
+          REPLACE(LOWER(title), ' ', '-') = ${normalizedTitle}
+          OR title = ${normalizedTitle}
+        )
+        AND (${isCreator} OR visibility != 'private')
+        LIMIT 1
+      `;
+
+      if (!topics[0]) return null;
+
+      // Fetch related data since raw query doesn't support include -- Added to match original behavior
       return await xprisma.topic.findFirst({
-        where: {
-          title: { equals: normalizedTitle, mode: "insensitive" },
-          creatorName: opts.input.username,
-          visibility: isCreator ? undefined : { not: "private" },
-        },
+        where: { id: topics[0].id },
         include: {
           nodes: true,
           edges: true,
--- a/src/pages/notifications.page.tsx
+++ b/src/pages/notifications.page.tsx
@@ -5,12 +5,11 @@ import { type MRT_ColumnDef, MaterialReactTable } from "material-react-table";
 import Head from "next/head";
 
 import { InAppNotification } from "@/common/inAppNotification";
-import { Topic } from "@/common/topic";
+import { Topic, normalizeTitle } from "@/common/topic";
 import { QueryError } from "@/web/common/components/Error/Error";
 import { Link } from "@/web/common/components/Link";
 import { Loading } from "@/web/common/components/Loading/Loading";
 import { useSessionUser } from "@/web/common/hooks";
-import { normalizeTitle } from "@/common/topic";
 import { trpc } from "@/web/common/trpc";
 
 type RowData = InAppNotification & { topic: Topic | null };
--- a/src/web/topic/components/TopicForm/TopicForm.tsx
+++ b/src/web/topic/components/TopicForm/TopicForm.tsx
@@ -23,7 +23,7 @@ import { useState } from "react";
 import { Controller, useForm } from "react-hook-form";
 import { z } from "zod";
 
-import { topicSchema, visibilityTypes, normalizeTitle } from "@/common/topic";
+import { normalizeTitle, topicSchema, visibilityTypes } from "@/common/topic";
 import { trpc } from "@/web/common/trpc";
 import { updateTopic as updateStoreTopic } from "@/web/topic/store/topicActions";
 import { generateBasicViews } from "@/web/view/quickViewStore/store";
@@ -50,7 +50,7 @@ export const CreateTopicForm = ({ creatorName }: { creatorName: string }) => {
   const onSubmit = (data: FormData) => {
     createTopic.mutate({
       topic: {
-        title: normalizeTitle(data.title),
+        title: data.title,
         description: data.description,
         visibility: data.visibility,
         allowAnyoneToEdit: data.allowAnyoneToEdit,
--- a/src/web/topic/components/TopicPane/TopicDetails.tsx
+++ b/src/web/topic/components/TopicPane/TopicDetails.tsx
@@ -27,7 +27,7 @@ import { useEffect, useState } from "react";
 import { useForm } from "react-hook-form";
 import { z } from "zod";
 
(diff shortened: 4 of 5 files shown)


lovelypangotra2003:
Hey @keyserj although not directly related to the PR I've found that when user enters a title it not only splits it based on hyphens but also camel case and displays in the text area to the user.
Example topic input : `AAbCd-EEEfgGh`
What user sees : `a ab cd ee efg gh`(split based on hyphen and camel case)
What user is expected to see : `aabcd eeefggh`(split based on just hyphen)
I don't know if this has been thought of.

lovelypangotra2003:
@keyserj take a look into this please

keyserj:
Started looking at this, will finish reviewing tomorrow

src/api/routers/topic.ts
```diff
@@ -22,14 +22,23 @@ export const topicRouter = router({
...
+      const normalizedTitle = normalizeTitle(decodeURIComponent(opts.input.title));
 
-      return await xprisma.topic.findFirst({
-        where: {
-          title: opts.input.title,
-          creatorName: opts.input.username,
-          visibility: isCreator ? undefined : { not: "private" },
-        },
-      });
+      const topics = await xprisma.$queryRaw<
```
keyserj, line 27: I didn't expect to have to turn most of the topic endpoints into raw queries/use the normalization checks (this is my lack of foresight). It's hard to say if this is the best path forward compared to the alternatives I mention in the below snippet, let me know if you have opinions.

I think the current strategy can be made a little better (and ok enough for me) by extracting a `findTopicViaUrlSafeTitle(username, title, isCreator)` method that normalizes the title, runs the raw query, and returns the topic if it exists. Then each route that's being updated in this PR can use that however it needs (return the result, or use the result's id to retrieve more fields via not-raw query, or throw an exception if we needed the result to be null for uniqueness).

Here are some thoughts that I'd like to go into a comment on the above method, to provide explanation/context for future maintainability:
```ts
/**
 * This is a bit of a hack to ensure that we treat spaces and hyphens as the same for topic titles.
 * We want this so that users can use either one, while ensuring that "topic-1" and "topic 1" can't
 * both exist, so that "topic 1" can use easier-to-read hyphens for URLs, e.g. "/user123/topic-1",
 * without conflict.
 *
 * Considered alternatives:
 * 1. allow spaces but just encode them in the URL e.g. "/user123/topic%201". this doesn't look very nice imo. probably way easier implementation though
 * 2. disallow hyphens in the title and only allow spaces, then convert spaces to hyphens for the URL. maybe this is good but hard to tell if disallowing hyphens is ok
 * 3. allow spaces and hyphens, but encode the spaces as `+` in the URL. this could look cleaner but it seems like `encodeURIComponent` actually encodes `+` too, so it's hard to tell if that's ok
 *
 * Note: some queries need different SELECT/JOINs, so our slightly-jank strategy to reuse this raw query logic
 * is to use this for finding, then run another query in order to grab other columns we need without
 * needing to use a raw query for that.
 *
 * Ideally:
 * - If we could use the query builder for everything and just the REPLACE expression was raw SQL,
 * then it'd be easy to use one query for any retrieval, rather than a separate query for this method
 * and a separate follow-up query for specific SELECTs.
 * Seems like there isn't an issue created for REPLACE support, but here's the issue for partial-raw support https://github.com/prisma/prisma/issues/5560
 * - If we could use a collation that treats spaces and hyphens as the same, then we could just use
 * the query builder for everything. But I couldn't find a collation that does this.
 */
```

src/api/routers/topic.ts
```diff
@@ -22,14 +22,23 @@ export const topicRouter = router({
...
+        { id: number; title: string; creatorName: string; visibility: string }[]
+      >`
+        SELECT id, title, "creatorName", visibility
```
keyserj, line 30:
```suggestion
        Topic
      >`
        SELECT *
```
In the shared method I mention above, I think it's fine to return all topic columns - then we can also use prisma's generated `Topic` type.

src/api/routers/topic.ts
```diff
@@ -22,14 +22,23 @@ export const topicRouter = router({
...
+        FROM "topics"
+        WHERE "creatorName" = ${opts.input.username}
+        AND (
+          REPLACE(LOWER(title), ' ', '-') = ${normalizedTitle}
+          OR title = ${normalizedTitle}
```
keyserj, line 35: It seems to me that, since our column is `citext`, we don't need `LOWER`, and since we're normalizing both the column and query parameter, we shouldn't need the OR that checks `title = ${normalizedTitle}`.

src/api/routers/topic.ts
```diff
@@ -41,18 +50,40 @@ export const topicRouter = router({
     .input(
       z.object({
         username: userSchema.shape.username,
-        title: topicSchema.shape.title,
+        title: z.string(),
```
keyserj, line 53: Hmm was this change intentional? It seems like we should still be using the zod schema to ensure our passed-in `title` is valid.

src/api/routers/topic.ts
```diff
@@ -22,14 +22,23 @@ export const topicRouter = router({
     )
     .query(async (opts) => {
       const isCreator = opts.input.username === opts.ctx.user?.username;
+      const normalizedTitle = normalizeTitle(decodeURIComponent(opts.input.title));
```
keyserj, line 25: Did you find it necessary to decode? It's my understanding that TRPC automatically encodes on the client-side and & decodes on the server side for us ([specification which seems to imply this](https://trpc.io/docs/rpc#methods---type-mapping)).
```suggestion
      const normalizedTitle = normalizeTitle(opts.input.title);
```

src/api/routers/topic.ts
```diff
@@ -268,6 +299,19 @@ export const topicRouter = router({
...
+      const normalizedTitle = normalizeTitle(opts.input.topic.title);
+      const existingTopic = await xprisma.topic.findFirst({
+        where: {
+          title: { equals: normalizedTitle, mode: "insensitive" },
+          creatorName: opts.ctx.user.username,
+        },
+      });
+      if (existingTopic) {
+        throw new TRPCError({
+          code: "BAD_REQUEST",
```
keyserj, line 311: I think `UNPROCESSABLE_CONTENT` or `CONFLICT` would be more accurate here https://stackoverflow.com/a/25016375/8409296

src/api/routers/topic.ts
```diff
@@ -41,18 +50,40 @@ export const topicRouter = router({
...
+          OR title = ${normalizedTitle}
+        )
+        AND (${isCreator} OR visibility != 'private')
+        LIMIT 1
+      `;
+
+      if (!topics[0]) return null;
 
+      // Fetch related data since raw query doesn't support include -- Added to match original behavior
       return await xprisma.topic.findFirst({
```
keyserj, line 85: `findUnique` is clearer/I think slightly more performant if we're querying by id.
```suggestion
      return await xprisma.topic.findUnique({
```

keyserj (commented):
This is a great start! I appreciate your work, this is a sizable effort.

> when user enters a title it not only splits it based on hyphens but also camel case and displays in the text area to the user

Can you screenshot this? I wouldn't expect it to happen, and I don't see it happening here:
image

Also note: after you open a PR (also when you commit to a PR), automated tests are run. At the bottom of the PR you can see that these actually failed, and you can click on them to see what failed:
image

In this case, the failures are because a topic title with uPPerAndLowerCasing can no longer be searched via URL with upper case. I believe the suggestions I made in other comments will resolve this failures.

keyserj:
Hey, sorry to throw you for a loop, but I was inspired to look further into collations, and I think I found an ideal solution here.

The article [Overview of ICU collation settings](https://peter.eisentraut.org/blog/2023/05/16/overview-of-icu-collation-settings) and the [postgres Collation Settings docs](https://www.postgresql.org/docs/current/collation.html#ICU-COLLATION-SETTINGS) give a nice explanation of ways to customize a collation. We can set the collation's locale to `und-u-ka-shifted-ks-level2` in order to ignore case _and_ punctuation + space characters during comparisons:
- `und`: I guess this uses some "good" default sorting settings?
- `u`: unicode (which is ok for our column even though we're just alphanumeric + space + hyphens)
- `ka-shifted`: ignore characters specified by `ks` setting during comparison
- `ks-level2`: ignore characters from levels 3 + 4 (see [table](https://www.postgresql.org/docs/current/collation.html#ICU-COLLATION-COMPARISON-LEVELS)) during comparison, i.e. ignore case + punctuation

Using this collation for our `title` column will mean that e.g. `topic 1` and `TOPIC-1` are treated as equal without modifying any queries at all, which seems to me like a much cleaner solution. I believe file changes aside from `routers.topic.ts` will still be needed. What do you think?

Note: creating a collation looks like [this](https://github.com/lovelypangotra2003/ameliorate/blob/81af0139b97225174ea17876b81b8f32c1b7872a/src/db/migrations/20240108134835_add_case_insensitive_collation/migration.sql#L4-L8) (and the migration after that one shows how to set the collation on a column).

lovelypangotra2003:
Hi @keyserj I'm sorry but I've to hold this for sometime as I am busy with something. I hope to return soon and take a look into this. I hope that'll be okay.

keyserj:
No worries, it's no rush. Thanks for letting me know.

keyserj:
Bump @lovelypangotra2003 do you still plan on coming back to this?

keyserj:
Going to close this for now, though we'll want to use these changes when we come back to it.

PR #722, opened 2025-05-15
Delete option on edge

SakshamKaundal:
#### Problem
Previously, the "Delete edge" option in the context menu was only available when right-clicking on the edge label or certain parts of the edge path. 

#### Solution
- Added the `onContextMenu` handler to all interactive edge elements, "NOT" including the hidden interaction path and the move/drag handle.
- Ensured the correct edge context is passed to the context menu, so the "Delete edge" option is always available when right-clicking any part of the edge.

#### Result
Users can now right-click anywhere on an edge—"NOT" including the move/drag handle—and always see the "Delete edge" option in the context menu, improving the editing workflow and consistency.

---

### DIFF
--- a/src/web/topic/components/Edge/ScoreEdge.tsx
+++ b/src/web/topic/components/Edge/ScoreEdge.tsx
@@ -122,6 +122,7 @@ export const ScoreEdge = ({ inReactFlow, ...flowEdge }: EdgeProps & Props) => {
       markerStart={`url(#${inReactFlow ? flowMarkerId : nonFlowMarkerId}-${spotlight})`}
       spotlight={spotlight}
       onClick={() => setSelected(edge.id)}
+      onContextMenu={(event) => openContextMenu(event, { edge })}
     />
   );
 
@@ -137,6 +138,10 @@ export const ScoreEdge = ({ inReactFlow, ...flowEdge }: EdgeProps & Props) => {
       strokeOpacity={0}
       strokeWidth={20}
       onClick={() => setSelected(edge.id)}
+      onContextMenu={(event) => {
+        console.log("Edge context menu:", edge);
+        openContextMenu(event, { edge });
+      }}
     />
   );
 
--- a/src/web/topic/components/Edge/ScoreEdge.tsx
+++ b/src/web/topic/components/Edge/ScoreEdge.tsx
@@ -138,10 +138,7 @@ export const ScoreEdge = ({ inReactFlow, ...flowEdge }: EdgeProps & Props) => {
       strokeOpacity={0}
       strokeWidth={20}
       onClick={() => setSelected(edge.id)}
-      onContextMenu={(event) => {
-        console.log("Edge context menu:", edge);
-        openContextMenu(event, { edge });
-      }}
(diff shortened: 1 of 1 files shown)


src/web/topic/components/Edge/ScoreEdge.tsx
```diff
@@ -137,6 +138,10 @@ export const ScoreEdge = ({ inReactFlow, ...flowEdge }: EdgeProps & Props) => {
       strokeOpacity={0}
       strokeWidth={20}
       onClick={() => setSelected(edge.id)}
+      onContextMenu={(event) => {
+        console.log("Edge context menu:", edge);
+        openContextMenu(event, { edge });
+      }}
```
keyserj, line 144: Shouldn't need this `console.log`
```suggestion
      onContextMenu={(event) => openContextMenu(event, { edge })}
```

src/web/topic/components/Edge/ScoreEdge.tsx
```diff
@@ -218,6 +223,7 @@ export const ScoreEdge = ({ inReactFlow, ...flowEdge }: EdgeProps & Props) => {
             height={100}
             style={{ position: "absolute", cursor: "default" }}
             className="react-flow__edge selected"
+            onContextMenu={(event) => openContextMenu(event, { edge })}
```
keyserj, line 226: What motivated you to add this here too? I would've expected your update to `path`'s `onContextMenu` to be sufficient for this spot, since `path` is rendered within this `svg`.

keyserj (commented):
Hey welcome to Ameliorate! 🙂 This is looking pretty good, just a couple small comments.

src/web/topic/components/Edge/ScoreEdge.tsx
SakshamKaundal: You are right...We don't need this here..sorry

SakshamKaundal: Why its showing commit is outdated?

src/web/topic/components/Edge/ScoreEdge.tsx
SakshamKaundal: he motivation for adding the onContextMenu handler to the <svg> element was to ensure that the context menu is triggered even when the user right-clicks on areas of the edge that are not directly covered by the path. While the path does have the onContextMenu handler, there are cases where users might right-click on empty parts of the svg (e.g., areas between the path and the label), and the event would not propagate correctly.

By adding the handler to the <svg>, we ensure a consistent user experience where right-clicking anywhere within the edge's bounding box triggers the context menu, regardless of whether the click lands directly on the path.

src/web/topic/components/Edge/ScoreEdge.tsx
keyserj: GitHub shows a comment as outdated if it's on a line of code that has been changed since the comment was made. I commented on the `console.log` which you've removed, so the comment is now outdated.

keyserj (approved):
This looks great 🚀 thanks for contributing!

PR #735, opened 2025-05-22
Text cut off in nodes

### LINKED ISSUE #557: Text is cut off in nodes
### Describe your issue

text

### Solution you'd like

There are multiple possible solutions and these could be discussed here.
* The nodes could be a little bit larger if there is more text in them and how large could be configurable (I'd suggest that the default would be quite a bit larger than now; this may either itself not fully solve this or one sets a max-length for these texts so that the text always fits into the node with any additional details going into the Notes of the item)
* The node could expand when hovering over it
* The nodes could become bigger and the font-size be decreased up to the max-length of the node text

However, I think for exporting a map, the full text of all items definitely needs to be visible. If you want to export an image version one can't hover over nodes. Nodes with long texts would need to be edited to improve how the problem tree map looks like when shared with others / exported but having a few somewhat larger nodes is much better than just making parts of the text not readable. So just showing the full text when hovering would not be sufficient for that...e.g. an export to image view would adjust the nodes sizes and then one could also have an option whether the full text should always be shown (I would enable that).

Maybe there are more ways this could be addressed.

_No response_

### Additional context

Mentioned in #535 where @keyserj wrote
> My intention was to keep node text small and concise so they can be quickly glanced at to understand high-level concepts. To this end, nodes are fixed size, and extra explanatory text can be included in the node's notes. This is open to discussion. It probably requires some practice for wording/sentence-fragmenting, but I've always been able to word nodes such that their concept fits in the current node's size.

Facilitating users keep texts in nodes short and succinct is good but there still needs to be a way to see the full text.

If text is entered longer than the max-length it should display an error and save the text up to the max length and ask the user  to shorten/summarize it and to put any additional info into the notes (Maybe that field was better be called "More information" or "More information and notes" or similar.)

_No response_

keyserj:
Closes #557 

### Description of changes

- see commit messages

### Additional context

-

### DIFF
--- a/src/web/topic/components/Node/EditableNode.styles.tsx
+++ b/src/web/topic/components/Node/EditableNode.styles.tsx
@@ -1,5 +1,5 @@
 import styled from "@emotion/styled";
-import { Box, TextareaAutosize } from "@mui/material";
+import { Box } from "@mui/material";
 
 import { htmlDefaultFontSize } from "@/pages/_document.page";
 import { ContentIndicatorGroup } from "@/web/topic/components/Indicator/Base/ContentIndicatorGroup";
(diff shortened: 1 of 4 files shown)


keyserj:
Demo below. At max reduction, four lines fit, which I think helps a lot. Going beyond the max reduction and four lines no longer has a scrollbar, mainly so font size calcs don't need to remove it to check fit if it exists, but I think it's fine because at this point there's very likely more text than necessary and I don't think it's worth supporting.

I plan to add a warning popup when font reduction happens for the user's first time, to warn about performance concerns when > 10s of displayed nodes have font reduced.

https://github.com/user-attachments/assets/1b764877-dcf3-4e5c-8afd-72551990c522

PR #737, opened 2025-05-29
touchup: use ellipses for quick view titles

### LINKED ISSUE #723: Use ellipses for quick view titles when they're too long
### Describe your issue

If I enter a quick view title that's even slightly long, it's pretty annoying to look at, since the overflow just goes behind buttons before getting cut off:

https://github.com/user-attachments/assets/74551b5c-38b4-4f66-95e2-28ff35b7d008

### Solution you'd like

We should be able to use ellipses when there's not enough space to display the full title. In the above video, the title could show "Breakdown super duper rea...".

_No response_

### Additional context

We probably should put the 3 quick view buttons (edit/save/delete) into a dropdown menu as well, to make more space for the title, but we can do that in a separate ticket.

- The [text-overflow css prop](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow) seems like it should accomplish what we want
- Code change should probably be in the [QuickViewSection](https://github.com/amelioro/ameliorate/blob/cdbefee32b849cc98fcc59460ff3a4bcff014877/src/web/topic/components/TopicPane/QuickViewSection.tsx#L0-L1) component

qdow:
Closes #723

### Description of changes

- Long quick view titles now use ellipses instead of overflowing behind the buttons

    Before:
before

    After:
after

- This change was implemented by adding `className="[&>span]:truncate"` to the `<ListItemText>` displaying the quick view title in the `QuickViewSection.tsx` file

### Additional context

- See the issue comments for a discussion of the change.

### DIFF
--- a/src/web/topic/components/TopicPane/QuickViewSection.tsx
+++ b/src/web/topic/components/TopicPane/QuickViewSection.tsx
@@ -90,7 +90,7 @@ const QuickViewRow = ({ view, selected, editable, onEdit }: RowProps) => {
           )}
         </ListItemIcon>
 
-        <ListItemText primary={view.title} />
+        <ListItemText primary={view.title} className="[&>span]:truncate" />
 
         {editable && (
           <>


keyserj:
@all-contributors please add @qdow for code

(let me know if you prefer not to be marked as a contributor in the project readme!)

qdow:
Yay! Thank you! 😄

PR #751, opened 2025-06-04
feat(info-dialog): add arrow to Popper tooltip

GarvThakral:
Closes #issue-number-here

### Description of changes
Fixes #740
-

### Additional context
No other files were changed . 
-

PR #753, opened 2025-06-05
Arrow it 2

### LINKED ISSUE #740: Add arrow to info dialog if it's anchored to an element
### Describe your issue

It's nice that we get a tip when typing too much text into a node, but the info dialog seems _slightly_ disconnected from the element it's being anchored to:

Image

Note: the tip only shows up once. To reproduce again, you can 1. clear local storage, 2. open in incognito, or 3. comment out [this line](https://github.com/amelioro/ameliorate/blob/323c8282b38584d016dfc72c3c55854519179315/src/web/topic/components/Node/NodeTextArea.tsx#L14) (if running locally), then type too much text into a node again.

### Solution you'd like

It seems like it'd help to have an arrow like some tooltips have:

Image

_No response_

### Additional context

_No response_

- [here's the file for our InfoDialog component](https://github.com/amelioro/ameliorate/blob/323c8282b38584d016dfc72c3c55854519179315/src/web/common/components/InfoDialog/InfoDialog.tsx)
- we're using [Material UI's Popper component](https://mui.com/material-ui/react-popper/) for this info dialog
- Material UI uses popperjs under the hood, [which has an arrow demo here](https://popper.js.org/docs/v2/modifiers/arrow/)
- [this example on stackoverflow](https://stackoverflow.com/a/64040075/8409296) seems like a good way of accomplishing this with our Material UI Popper

GarvThakral:
Closes #740 

### Description of changes
Added arrow functionality . 
Added back the older comments
Further comments have been added for better readability
Added appropriate colors .
### Additional context
Since we cannot move the anchor while the dialog box is opened , I am unable to reach a case where the popper would appear on the left or right . But for the top and bottom its works way better than the last commit  .

### DIFF
--- a/src/web/common/components/InfoDialog/InfoDialog.tsx
+++ b/src/web/common/components/InfoDialog/InfoDialog.tsx
@@ -23,10 +23,11 @@ export const InfoDialog = () => {
   const [infoType, setInfoType] = useState<InfoType>("info");
   const [message, setMessage] = useState<ReactNode>("");
   const [anchor, setAnchor] = useState<Anchor | null>(null);
+  const [arrowRef, setArrowRef] = useState<HTMLElement | null>(null);
 
   useEffect(() => {
     // seems a little awkward to rely on an event to set all this state, but what's nice about it is
-    // that the state doesn't need to be accesssible from outside the component.
+    // that the state doesn't need to be accessible from outside the component.
     const unbindInfoToShow = emitter.on("infoToShow", (infoType, message, anchor) => {
       setInfoType(infoType);
       setMessage(message);
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
       <Popper
         open={open}
         anchorEl={anchorEl}
-        modifiers={[{ name: "offset", options: { offset: [0, 12] } }]}
-        // z-index to match the mask's so that popper is above the mask
+        placement="bottom"
+        modifiers={[
+          // allow Popper to flip between bottom and top if there's not enough space
+          { name: "flip", options: { fallbackPlacements: ["bottom", "top"] } },
+          // push Popper down/up by 26px from the anchor
+          { name: "offset", options: { offset: [0, 26] } },
+          // enable arrow positioning, using our arrowRef
+          { name: "arrow", options: { element: arrowRef, padding: 5 } },
+          {
+            name: "preventOverflow",
+            options: {
+              altAxis: true,
+              altBoundary: true,
+              tether: true,
+              rootBoundary: "document",
+              padding: 12,
+            },
+          },
+        ]}
+        // z-index to match the mask's so that Popper is above the mask
         className={
-          "z-[99999] max-w-[min(80svw,36rem)] rounded-lg bg-white" +
+          "overflow-visible z-[99999] max-w-[min(80svw,36rem)] rounded-lg bg-white" +
           (anchorEl ? "" : " !left-1/2 !top-1/2 !-translate-x-1/2 !-translate-y-1/2")
         }
         aria-label="Info Dialog"
       >
-        <div className="flex items-center justify-between rounded-t-lg bg-paperShaded-main px-2 py-1">
-          <InfoTypeIcon color={infoType} fontSize="small" />
+        {({ placement }) => {
+          const isBelow = placement.startsWith("bottom");
 
-          <Typography className="font-bold">{startCase(infoType)}</Typography>
+          return (
+            <>
+              <span
+                ref={setArrowRef}
+                data-popper-arrow
+                className={`
+                  absolute
+                  w-3
+                  h-3
+                  overflow-visible
+                  ${isBelow ? "top-[-6px]" : "bottom-[-6px]"}
+                  before:content-['']
+                  before:absolute
+                  before:top-1/2
+                  before:left-1/2
+                  before:w-3
+                  before:h-3
+                  ${isBelow ? "before:bg-paperShaded-main" : "before:bg-white"}
+                  before:translate-x-[-50%]
+                  before:translate-y-[-50%]
+                  before:rotate-45
+                `}
+                style={{ zIndex: 99999 }}
+              />
 
-          <IconButton size="small" title="Close" aria-label="Close" onClick={() => setOpen(false)}>
-            <Close fontSize="small" />
-          </IconButton>
-        </div>
+              <div className="flex items-center justify-between rounded-t-lg bg-paperShaded-main px-2 py-1">
+                <InfoTypeIcon color={infoType} fontSize="small" />
 
-        <Typography variant="body2" className="max-h-[70svh] overflow-auto whitespace-pre-wrap p-2">
-          {message}
-        </Typography>
+                <Typography className="font-bold">{startCase(infoType)}</Typography>
+
+                <IconButton
+                  size="small"
+                  title="Close"
+                  aria-label="Close"
+                  onClick={() => setOpen(false)}
+                >
+                  <Close fontSize="small" />
+                </IconButton>
+              </div>
+
+              <Typography
+                variant="body2"
+                className="max-h-[70svh] overflow-auto whitespace-pre-wrap p-2"
+              >
+                {message}
+              </Typography>
+            </>
+          );
+        }}
       </Popper>
     </>
   );
--- a/src/web/common/components/InfoDialog/InfoDialog.tsx
+++ b/src/web/common/components/InfoDialog/InfoDialog.tsx
@@ -68,7 +68,7 @@ export const InfoDialog = () => {
           // push Popper down/up by 26px from the anchor
           { name: "offset", options: { offset: [0, 26] } },
           // enable arrow positioning, using our arrowRef
-          { name: "arrow", options: { element: arrowRef, padding: 5 } },
+          { name: "arrow", options: { element: arrowRef, padding: 10 } },
           {
             name: "preventOverflow",
             options: {
--- a/src/web/common/components/InfoDialog/InfoDialog.tsx
+++ b/src/web/common/components/InfoDialog/InfoDialog.tsx
@@ -66,16 +66,6 @@ export const InfoDialog = () => {
           { name: "offset", options: { offset: [0, 26] } },
           // enable arrow positioning, using our arrowRef
           { name: "arrow", options: { element: arrowRef, padding: 10 } },
-          {
-            name: "preventOverflow",
-            options: {
-              altAxis: true,
-              altBoundary: true,
-              tether: true,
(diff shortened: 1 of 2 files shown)


src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
...
-        modifiers={[{ name: "offset", options: { offset: [0, 12] } }]}
-        // z-index to match the mask's so that popper is above the mask
+        placement="bottom"
+        modifiers={[
+          // allow Popper to flip between bottom and top if there's not enough space
+          { name: "flip", options: { fallbackPlacements: ["bottom", "top", "left"] } },
+          // push Popper down/up by 26px from the anchor
+          { name: "offset", options: { offset: [0, 26] } },
+          // enable arrow positioning, using our arrowRef
+          { name: "arrow", options: { element: arrowRef, padding: 10 } },
```
keyserj, line 71: Is this `padding` doing anything? It seems like the `offset` covers this

src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
...
-          <Typography className="font-bold">{startCase(infoType)}</Typography>
+          return (
+            <>
+              <span
+                ref={setArrowRef}
+                data-popper-arrow
+                className={`
+                  absolute
+                  w-3
+                  h-3
```
keyserj, line 101: Small tailwind nitpicks: these can be `size-3`, and `top-[-6px]` can be `-top-1.5`. There's a tailwind linter that should catch these simplifications. I wonder if you might not have the linter set up? Check out [this comment](https://github.com/amelioro/ameliorate/pull/580#discussion_r1864994128) which describes setting it up.

src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
...
+                  overflow-visible
```
keyserj, line 102: As above, I'd expect `overflow-visible` to be undesirable here

src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
...
+                  ${isBelow ? "top-[-6px]" : "bottom-[-6px]"}
+                  before:content-['']
+                  before:absolute
+                  before:top-1/2
+                  before:left-1/2
```
keyserj, line 107: These `before:top-1/2`, `before:left-1/2`, `before:translate-x-[-50%]`, `before:translate-y-[-50%]` actually seem to be covered by popper's default positioning of the `data-popper-arrow`

src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
...
+                  before:w-3
+                  before:h-3
+                  ${isBelow ? "before:bg-paperShaded-main" : "before:bg-white"}
+                  before:translate-x-[-50%]
+                  before:translate-y-[-50%]
+                  before:rotate-45
+                `}
+                style={{ zIndex: 99999 }}
```
keyserj, line 115: Did you find this z-index necessary? It seems like this span should already be within the parent Popper's stacking context, which would put it in front of everything

src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
...
       >
-        <div className="flex items-center justify-between rounded-t-lg bg-paperShaded-main px-2 py-1">
-          <InfoTypeIcon color={infoType} fontSize="small" />
+        {({ placement }) => {
+          const isBelow = placement.startsWith("bottom");
 
-          <Typography className="font-bold">{startCase(infoType)}</Typography>
+          return (
+            <>
+              <span
```
keyserj, line 95: Since the last PR, it looks like you've removed functionality to only show the arrow when there's an anchorEl. We should keep that in

keyserj (commented):
It's looking better!

In the future, please keep to one PR per issue. By creating a second PR, there's lost context from the comments in the first PR. When you want to address feedback, you should be able to add or modify commits on the same branch you first used, so that all the context for the changes are kept in one PR. I'm happy to provide guidance with git if that's the issue here.

For context, here is a link to the previous PR #751

src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
       <Popper
         open={open}
         anchorEl={anchorEl}
-        modifiers={[{ name: "offset", options: { offset: [0, 12] } }]}
-        // z-index to match the mask's so that popper is above the mask
+        placement="bottom"
+        modifiers={[
+          // allow Popper to flip between bottom and top if there's not enough space
+          { name: "flip", options: { fallbackPlacements: ["bottom", "top", "left"] } },
```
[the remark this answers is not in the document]
GarvThakral: actually , the arrow was not coming out of the div and wasn't visible so a lot of code was me trying out things from the docs to make sure the functionality is robust . But on another review , i can see it isn't really necessary .

src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
...
+              tether: true,
+              rootBoundary: "document",
+              padding: 12,
+            },
+          },
+        ]}
+        // z-index to match the mask's so that Popper is above the mask
         className={
-          "z-[99999] max-w-[min(80svw,36rem)] rounded-lg bg-white" +
+          "overflow-visible z-[99999] max-w-[min(80svw,36rem)] rounded-lg bg-white" +
```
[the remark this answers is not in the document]
GarvThakral: overflow-visible , again was added so that the arrow would leave the info div , i will check for the functionality again and remove if its unnecessary .

src/web/common/components/InfoDialog/InfoDialog.tsx
GarvThakral: The padding keeps the arrow away from the rounded corners of the dialog , since the corners aren't sharp , there is a clear division when the arrow is at the extremes .

src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,84 @@ export const InfoDialog = () => {
...
+        placement="bottom"
+        modifiers={[
+          // allow Popper to flip between bottom and top if there's not enough space
+          { name: "flip", options: { fallbackPlacements: ["bottom", "top", "left"] } },
+          // push Popper down/up by 26px from the anchor
+          { name: "offset", options: { offset: [0, 26] } },
+          // enable arrow positioning, using our arrowRef
+          { name: "arrow", options: { element: arrowRef, padding: 10 } },
+          {
+            name: "preventOverflow",
```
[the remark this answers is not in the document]
GarvThakral: Screenshot From 2025-06-06 00-00-48
preventOverflow allows for something like this . 
Screenshot From 2025-06-06 00-02-15
This is what we see without this modifier . 

I will edit it however you suggest , personally i find the functionality with preventOverflow to be a bit better , more responsive .

src/web/common/components/InfoDialog/InfoDialog.tsx
GarvThakral: Screenshot From 2025-06-06 00-04-07
This is without the padding

keyserj:
> I would actually appreciate it if you can guide me a little
on making a pr and what are the best practices .

It'll be easiest to provide guidance if you have specific issues you're running into, but here are a few tips:
1. You can see which branch you used for the PR here
image
2. You can see which branch you're currently on via `git status` (`git checkout [branch name]` if you're not on the right branch)
3. After receiving feedback, you should be able to stay on the same branch and add new commits on top of what you have `git add .` `git commit -m "message"` `git push`.
4. If you want to practice keeping a clean git history you can check out the rebase & interactive rebase videos that I link [in this comment](https://github.com/amelioro/ameliorate/pull/580#discussion_r1864992693) - but I will be squashing extra commits before merging anyway, so you don't have to do this.
5. There are some further tips here https://opensource.guide/how-to-contribute/#opening-a-pull-request though I think you did most of those things pretty well

src/web/common/components/InfoDialog/InfoDialog.tsx
keyserj: Oooh great eye, thanks for the image. That makes sense

src/web/common/components/InfoDialog/InfoDialog.tsx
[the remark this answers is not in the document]
keyserj: Ah that makes sense, the images help a lot, thanks for that.

I think
1. generally the element being interacted with will not be partially off-screen
2. the element can be behind a pane but even in this case, while the arrow doesn't line up with the part of the element that's visible, it still lines up with the highlighted section

and because of these, it feels like specifying the extra options has a cost without much value. 

So I think I'd slightly prefer not having the options, but honestly I think it would be reasonable to keep as well. I'll leave it up to you.

src/web/common/components/InfoDialog/InfoDialog.tsx
```diff
@@ -60,27 +61,76 @@ export const InfoDialog = () => {
...
+                  ref={setArrowRef}
+                  data-popper-arrow
+                  className={`
+                  absolute
+                  size-3
+                  ${isBelow ? "-top-1.5" : "-bottom-1.5"}
+                  before:content-['']
+                  before:absolute
+                  before:w-3
+                  before:h-3
```
keyserj, line 103: This can also be simplified to `before:size-3`
```suggestion
                  before:size-3
```

src/web/common/components/InfoDialog/InfoDialog.tsx
keyserj: I think you ended up adding a commit, which is the right thing to do because that's the easiest way for me to see and review the changes anyway. They won't affect production until I merge them in any case

keyserj (commented):
Good work on this, I appreciate your explanations for why you specified the various options. Just a few more small comments

keyserj:
@all-contributors please add @GarvThakral for code

(let me know if you prefer not to be marked as a contributor in the project readme!)

PR #859, opened 2026-01-16
feat:added keyboard shortcuts for undo and redo

### LINKED ISSUE #707: Hotkey for undo/redo
### Describe your issue

It's nice that we've got undo/redo buttons in the Header:

Image

but if my hands are on the keyboard, I lose hundreds of milliseconds moving my hands to the mouse in order to click one of those options.

### Solution you'd like

If there's something to undo, pressing Ctrl + Z should undo it.
If there's something to redo, pressing Ctrl + Shift + Z should redo it.

_No response_

### Additional context

_No response_

Should be very similar to [setting the score for the selected node/edge](https://github.com/amelioro/ameliorate/blob/a9080cee51cb396313653cb4e007388bb62d5bc3/src/web/topic/components/TopicWorkspace/TopicWorkspace.tsx#L29-L39), except:
- instead of doing something with the score, we need to [perform undo or redo](https://github.com/amelioro/ameliorate/blob/c5d3981db11418d190647ab919a174b80626a379/src/web/topic/store/utilActions.ts#L62-L68)
- also probably check that undo/redo is possible, should be able to [check if there are past/future states available like this but without a hook](https://github.com/amelioro/ameliorate/blob/c5d3981db11418d190647ab919a174b80626a379/src/web/topic/store/utilHooks.ts#L11-L12)
- we should also only continue with undo/redo if the user has permission to; should be able to check via the [canEdit hook](https://github.com/amelioro/ameliorate/blob/a9080cee51cb396313653cb4e007388bb62d5bc3/src/web/topic/components/Node/EditableNode.tsx#L33-L34)

sudeepchalla:
Resolves #707 

### Description of changes

- Added undo and redo definitions in hotkeys.ts
- Added the undo/redo keys in useWorkspaceHotKeys in TopicWorkspace.tsx
  - Includes checks for canUndo/canRedo and userCanEditTopicData

### Additional context

- Added default event.preventDefault() to fix:
  - Canvas actions (like adding a node, editing, etc) and the text in comments/description being simulateously undo/redo

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -78,6 +78,15 @@ export default defineConfig(
             "index",
             "unknown",
           ],
+          // Ensure @/ and ~/ path aliases are treated as "internal" group, even if the TypeScript
+          // resolver fails to resolve them (which would otherwise classify them as "unknown").
+          // Unsure why resolver would fail, since npm/eslint/ts versions should all be pinned via
+          // package.json, but it seems to have happened for a contributor here https://github.com/amelioro/ameliorate/pull/859
+          // resulting in eslint reordering imports in a different way.
+          pathGroups: [
+            { pattern: "@/**", group: "internal" },
+            { pattern: "~/**", group: "internal" },
+          ],
           "newlines-between": "always",
           alphabetize: { order: "asc" },
         },
--- a/scripts/seed.ts
+++ b/scripts/seed.ts
@@ -2,12 +2,11 @@
 
 import { v4 as uuid } from "uuid";
 
+import { Edge, Node } from "@/db/generated/prisma/client";
 
 import { RelationName } from "../src/common/edge";
 import { NodeType } from "../src/common/node";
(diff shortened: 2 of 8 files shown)


sudeepchalla:
@keyserj 
Hey, I've added the comments, changed the casing in hotkey.
On running eslint fix it changed a bunch of files, is it okay i assume?

keyserj:
Hey @sudeepchalla your changes look good except the additional eslint impacted files. I tried looking into what may have caused your `npx eslint . --fix` to order the imports differently (it placed aliased `@` imports below relative imports `../..`), and alas I couldn't quite figure it out. I'd suspect that perhaps differing eslint or ts or npm versions could cause it, but the project's `package.json` should pin these to the correct versions when you `npm install`.

I updated the eslint config to explicitly classify `@` imports as "internal" imports (which is how they should be classified). Can you try `git pull` on your branch to get the commit I added and then run `npx eslint . --fix` again to see if it still changes more files?

PR #882, opened 2026-02-19
fix: show navbar and clearer error for unauthenticated private topics

### LINKED ISSUE #877: Visiting a private topic without permission only shows error without navbar or guidance
### Describe the bug

If I make a private topic, then log out (or open incognito browser) and try to visit the page, all I see is a 404 and nothing else:

Image

This is very confusing if the user doesn't realize they were logged out, and it's annoying that there are no links (or navbar) to get back to Ameliorate.

### To Reproduce

_No response_

### Expected behavior

1. show the navbar
2. error should be more explicit. e.g. "topic not found. either the topic doesn't exist, or you don't have permission to view it. `Home | Login`". (login button if you're not logged in)

### Additional context

_No response_

### Technical ideas

_No response_

rishabhrthr001:
Closes #877

This change standardizes error rendering across the application and fixes
header visibility issues when errors occur outside the workspace layout.

Key changes:
- Introduced a shared AppError component to unify layout, spacing, and
  typography across all error states while allowing flexible actions via
  composition.
- Refactored existing error pages (topic access, query, auth, not found,
  and root errors) to use the shared AppError abstraction.
- Removed route-based conditional rendering from SiteHeader so it is always
  mounted, ensuring navigation does not disappear when error boundaries
  short-circuit normal rendering.
- Added a global CSS rule to hide the SiteHeader only when the workspace
  successfully mounts, keeping layout behavior declarative and error-safe.
- Removed misuse of Next.js <Head> for layout control and deleted the
  one-off TopicAccessError implementation.

Note:
Local pre-commit was blocked by an interactive `next lint` prompt; CI lint
will validate these changes.

I’ve implemented all the requested changes.

Summary:
Renamed TopicAccessError to TopicNotFoundError to better reflect missing topics rather than auth issues, and updated the messaging accordingly.
Updated AppError defaults to show Go Home always and Login only when the user is logged out, while preserving children overrides.
Moved the header visibility logic fully into Tailwind using [body:has(#workspace)_&]:hidden, removing the global CSS and header id.

Verification:
TypeScript checks pass (npx tsc --noEmit).
Manually verified topic behavior:
Logged out → Login + Home buttons
Logged in → Home button only
Workspace correctly hides the header.

Let me know if you’d like any additional tweaks.

### DIFF
--- a/src/web/common/components/Error/Error.tsx
+++ b/src/web/common/components/Error/Error.tsx
@@ -3,6 +3,7 @@ import * as Sentry from "@sentry/nextjs";
 import { ReactNode } from "react";
 
 import { Link } from "@/web/common/components/Link";
+import { useSessionUser } from "@/web/common/hooks";
 
 const logWarning = (message: string, statusCode: number) => {
   Sentry.setTag("statusCode", statusCode);
@@ -27,6 +28,9 @@ export const AppError = ({
   message?: ReactNode;
   children?: ReactNode;
 }) => {
+  const { sessionUser } = useSessionUser();
+  const isLoggedIn = sessionUser != null;
+
   return (
     <div className="mx-auto flex max-w-2xl flex-col items-center gap-4 px-4 py-24 text-center">
       <Typography variant="h4" component="h1" fontWeight="bold">
@@ -38,9 +42,16 @@ export const AppError = ({
 
       <div className="mt-4 flex justify-center gap-4">
         {children ?? (
-          <Button component={Link} href="/" variant="contained">
-            Go Home
-          </Button>
+          <>
+            <Button component={Link} href="/" variant="contained">
+              Go Home
+            </Button>
+            {!isLoggedIn && (
+              <Button component={Link} href="/api/auth/login" variant="outlined">
+                Login
+              </Button>
+            )}
+          </>
         )}
       </div>
     </div>
@@ -67,15 +78,11 @@ export const NotLoggedInError = () => {
   return <AppError statusCode={401} title="You must be logged in to view this page" />;
 };
 
-export const TopicAccessError = () => {
+export const TopicNotFoundError = () => {
   return (
     <AppError
       title="Topic not found"
       message="Either this topic doesn’t exist, or you don’t have permission to view it."
-    >
-      <Button component={Link} href="/" variant="contained">
-        Go Home
-      </Button>
-    </AppError>
+    />
   );
 };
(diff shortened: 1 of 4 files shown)


keyserj:
@rishabhrthr001 if you have no preference, feel free to just do it all on this PR.

src/web/common/components/Error/Error.tsx
```diff
@@ -13,22 +16,66 @@ interface Error {
...
+
+export const TopicAccessError = () => {
+  return (
+    <AppError
+      title="Topic not found"
+      message="Either this topic doesn’t exist, or you don’t have permission to view it."
+    >
+      <Button component={Link} href="/" variant="contained">
+        Go Home
+      </Button>
```
keyserj: Here it would be nice to have both the Go Home _and_ the Login button (if the user isn't logged in)

src/web/common/components/Error/Error.tsx
```diff
@@ -13,22 +16,66 @@ interface Error {
...
+  children?: ReactNode;
+}) => {
+  return (
+    <div className="mx-auto flex max-w-2xl flex-col items-center gap-4 px-4 py-24 text-center">
+      <Typography variant="h4" component="h1" fontWeight="bold">
+        {statusCode ? `${statusCode} - ` : ""}
+        {title ?? "An error occurred"}
+      </Typography>
+
+      {message && <Typography color="text.secondary">{message}</Typography>}
```
keyserj: TIL you can set MUI colors via `a.b` format, nice

src/web/common/components/Error/Error.tsx
```diff
@@ -13,22 +17,67 @@ interface Error {
...
 
 export const NotLoggedInError = () => {
   logWarning("NotLoggedInError rendered", 401);
 
-  return <NextError statusCode={401} title="You must be logged in to view this page" />;
+  return <AppError statusCode={401} title="You must be logged in to view this page" />;
+};
+
+export const TopicNotFoundError = () => {
+  return <AppError title="Topic not found" message="This topic doesn’t exist." />;
```
keyserj: This message should still be vague, since we might not be finding the topic due to lack of privilege

```suggestion
  return <AppError title="Topic not found" message="Either this topic doesn’t exist, or you don’t have permission to view it." />;
```

rishabhrthr001:
Thanks a lot for the thoughtful review and guidance — really appreciate it.
Learned a bunch working through this. Glad it was helpful!

PR #886, opened 2026-02-22
fix: collapse topic card after save  (#885)

### LINKED ISSUE #885: Save button should collapse topic card after saving
### Describe your issue

Currently, when editing a topic and pressing Save, the topic card stays open.
It would improve UX if the card automatically collapses after saving.

Steps to reproduce:
1. Open any topic (e.g. ameliorate.app/naughty8020/sample)
2. Edit the topic details
3. Press Save
4. Observe that the card stays open

Expected behavior:
- After pressing Save, the topic card collapses automatically.

This is a small QoL improvement for smoother interaction.

### Solution you'd like

When I press the save button, I would like the card to automatically collapse. 

_No response_

### Additional context

_No response_

Currently, the card collapses when clicking outside of it. A possible solution is to call this same collapse logic automatically after the save operation completes, so the card closes immediately without the user needing to click outside.

Naughty8020:
Closes #issue-number-here
Closes #885

### Description of changes

- Collapse topic card after saving 

### Additional context

- The flicker may sometimes occur due to the dialog’s rendering process.
- Looking for suggestions on how to reduce this flicker further.

### DIFF
--- a/src/web/topic/components/TopicForm/TopicForm.tsx
+++ b/src/web/topic/components/TopicForm/TopicForm.tsx
@@ -84,10 +84,9 @@ export const EditTopicForm = ({ topic, creatorName, afterSave }: EditTopicFormPr
   const [deleteDialogOpen, setDeleteDialogOpen] = useState(false);
 
   const updateTopic = trpc.topic.update.useMutation({
-    
     onSuccess: (updatedTopic) => {
       // need to update the URL if we're changing from the topic's page in the workspace
-      
+
       if (afterSave) afterSave();
       const url = new URL(window.location.href);
       const oldPath = `/${creatorName}/${topic.title}`;


src/web/topic/components/TopicForm/TopicForm.tsx
```diff
@@ -73,7 +74,7 @@ export const CreateTopicForm = ({ creatorName }: { creatorName: string }) => {
   return <TopicForm creatorName={creatorName} onSubmit={onSubmit} />;
 };
 
-export const EditTopicForm = ({ topic, creatorName }: { topic: Topic; creatorName: string }) => {
+export const EditTopicForm = ({ topic, creatorName,afterSubmit }: { topic: Topic; creatorName: string;afterSubmit?: () => void;}) => {
```
keyserj: Also, when our parameter list gets long, we usually create a `Props` interface to keep the function prototype a bit easier to read. Can you make an interface like this? You probably can call it `EditTopicFormProps`. Here's an example https://github.com/Naughty8020/ameliorate/blob/c31e269699699eac4f19a3afda35c6c0fca2e159/src/web/topic/components/TopicForm/TopicForm.tsx#L243-L250

src/pages/[username].page.tsx
```diff
@@ -130,7 +130,7 @@ const User: NextPage = () => {
               onClose={() => setEditingTopicId(null)}
               aria-label="Topic Settings"
             >
-              <EditTopicForm topic={row.original} creatorName={foundUser.username} />
+              <EditTopicForm topic={row.original} creatorName={foundUser.username}  afterSubmit={() => setEditingTopicId(null)} />
```
Comment (keyserj) [inline src/pages/[username].page.tsx]: > MyTopics may sometimes flicker briefly when saving

Can you explain a bit more? I'm not sure what you mean by this. Here is what I see when saving - it doesn't seem like My Topics is flickering:

https://github.com/user-attachments/assets/7c807969-a4e0-420a-9aab-15c676a906c8

https://github.com/user-attachments/assets/6aa96a18-4be6-40ad-be45-5751c66c154a

Naughty8020:
Good point — I can update the button label to “Save & Close” 👍  
I’ll push a small commit for that.

Thanks for digging into the flicker reproduction

Naughty8020:
You're right — I can reproduce it too when the topic being edited is not at the top of the list.

Naughty8020:
Thanks for the suggestion — that makes sense.
I’ve updated the button label to “Save & Close” as suggested.

src/pages/[username].page.tsx
```diff
@@ -130,7 +130,11 @@ const User: NextPage = () => {
               onClose={() => setEditingTopicId(null)}
               aria-label="Topic Settings"
             >
-              <EditTopicForm topic={row.original} creatorName={foundUser.username} />
+              <EditTopicForm
+                topic={row.original}
+                creatorName={foundUser.username}
+                afterSave={() => setEditingTopicId(null)}
+              />
             </Dialog>
```
Comment (keyserj) [inline src/pages/[username].page.tsx]: Ok I think I've figured out the flickering issue. If you look a few lines above this, we're actually rendering this dialog _per row_ of the table. I think what's happening is:
1. our `onSuccess` runs [code to update the user's cached topics](https://github.com/Naughty8020/ameliorate/blob/6b932924d00918128493cb72baa7bd33d4d36896/src/web/topic/components/TopicForm/TopicForm.tsx#L107-L119), which will trigger re-render of our table rows
2. the not-first row renders with a different topic id now that rows have reordered, so `editingTopicId === row.original.id` is false, making the Dialog disappear
3. the first row renders with the edited topic id so `editingTopicId === row.original.id` is now true for the first row, making the Dialog appear
4. our `afterSubmit` runs at the end of the `onSuccess`, finally clearing our `editingTopicId` and making the Dialog disappear again

I think we should just move our `Dialog` _outside_ of the `renderRowActions` so that we only ever render one editing `Dialog`, when `editingTopic` is not null (we'll have to create an `editingTopic` variable that's set based on looking through `rowData` for the `editingTopicId`, so we can pass that into the form rather than `row.original`). The `Dialog` can live above or below the `MaterialReactTable` component.

Another solution is to move the `afterSubmit` to the topic of the `onSuccess`, so that `editingTopicId` is cleared. We could do this additionally, but I think we should at least do the `Dialog` moving solution because it makes sense to only render one `Dialog`.

What do you think?

Naughty8020:
Thanks for the reply!  
I was able to fix the flickering by moving `afterSave` to the start of `onSuccess`.

PR #899, opened 2026-03-03
feat: added hotkey for deleting selected node/edge

### LINKED ISSUE #520: Hotkey for delete selected node/edge

henrybarl0w:
Closes #520

### Description of changes

- Added delete definition in hotkeys.ts
- Added the delete key in useWorkspaceHotKeys in TopicWorkspace.tsx
  - With check that user has permission using the canEdit hook

### Additional context

- When typing in a textfield the delete key performs its normal function

### DIFF
--- a/prisma.config.ts
+++ b/prisma.config.ts
@@ -15,5 +15,9 @@ export default defineConfig({
     // Also: using process.env with fallback instead of env() helper so that `prisma generate`
     // works in CI without a real database URL (since it doesn't actually connect to the DB).
     url: process.env.DIRECT_URL ?? "",
+    // should only be set during `generateMigration` because we need to handle the shadow db
+    // manually for the diffing that the script does. otherwise this should be undefined so that
+    // prisma can handle the shadow db.
+    shadowDatabaseUrl: process.env.SHADOW_DATABASE_URL,
   },
 });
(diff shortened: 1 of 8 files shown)


keyserj (approved):
This looks solid 🚀 thanks for your contribution!

PR #905, opened 2026-03-10
feat: turn on "all" comment notifications after editing a topic, if no preference set

### LINKED ISSUE #731: Turn on "all" comment notifications after editing a topic, if I haven't set a preference
### Describe your issue

If I've edited someone else's topic, it's annoying when I don't get notifications for future comments, which may be based on edits I've made. I can manually turn on "all" notifications after editing, but I don't always think of it.

### Solution you'd like

Turn on "all" comment notifications after editing a topic, if I haven't set a preference yet.

_No response_

### Additional context

- [Discord thread](https://discord.com/channels/1057707973482401892/1372272588600115330/1372272922034573513) with realization of some notifications being missed

_No response_

henrybarl0w:
Closes #731

### Description of changes

- When updates to the diagram are made though the updateDiagram endpoint in src/api/routers/topic.ts, check to see if the graph has been updated and then call handleGraphPartChanged function
- Added handleGraphPartChanged which checks for existing watch, if null is returned then creates a new watch of type "all"

### Additional context

- I have written the handleGraphPartChanged function in a new file located in src/api/notifications/watch.ts though as I am fairly unfamiliar with the codebase this may not be where it should be


src/api/notifications/watch.ts
```diff
@@ -0,0 +1,16 @@
+import { xprisma } from "@/db/extendedPrisma";
+
+export const handleGraphPartChanged = async (username: string, id: number) => {
```
keyserj: Can clarify these param names a bit
```suggestion
export const handleGraphPartChanged = async (editorUsername: string, topicId: number) => {
```

PR #915, opened 2026-03-24
feat: add click-to-zoom for tutorial images/gifs

### LINKED ISSUE #491: Add click-to-zoom for tutorial images/gifs
### Describe your issue

The images and gifs in the tutorial are pretty small and sometimes hard to see.

### Solution you'd like

It would be nice if I could easily click on the image/gif to maximize it to its full size. Like Miro:
chrome_2024-08-22_10-35-18

Adding the step's other info (e.g. title/text) to the modal could be good too, but not necessary.

_No response_

### Additional context

_No response_

Probably involves setting the cursor to the magnifying glass on hover, and opening a [modal](https://mui.com/material-ui/react-modal/) that takes up most of the screen

henrybarl0w:
Closes #491

### Description of changes

- Added a new ZoomableImage component.
- Creates a Modal when the image is clicked on with provide title and text.
- Updated all existing tutorial images to use ZoomableImage.
Screenshot 2026-03-24 at 13 16 59

### Additional context

- Images are only increased to their orginal file size so small images aren't enlarged.
- Incredibly large images may exeded the edge of the viewport but none of the current tutorial images do.
- Can be funny with thin aspect ratios.


src/web/tutorial/steps/addingNuance.tsx
```diff
@@ -27,14 +28,28 @@ export const addingNuanceSteps: StepType[] = [
...
+              <span>
+                You can use scores to convey positive or negative opinion about nodes and edges.
+                Click the "Show indicators" button <TabUnselected /> in the toolbar to show scores
+                and other indicators.
+                <br />
+                <br />
+                Here we're saying that we think pedestrians getting hit is a big concern, and that
+                people getting places faster is a benefit but we don't care about it at all.
+              </span>
+            }
```
keyserj: I think it's a bit too much to duplicate the title and text here for every step that uses an image like this. Though it seems a bit tricky to avoid this duplication.

Here are the ideas I can think of to avoid this:
1. don't add title or text to the modal. this is the easiest but probably not the friendliest
2. don't actually pass a `ZoomableImage` in - keep the same exact `imageSlot` prop, put the magnifying glass cursor CSS and zoom `onClick` onto the `StepContent`'s `div` that surrounds the `imageSlot`, then the `StepContent` can create the modal and fill it with its own title/text. this could be awkward for the cases where we pass more than just a single image into `imageSlot`, since _everything_ in that slot would have the hover/click. AFAIK we have some situations where we pass captions, and some cases where we pass two images. we might be able to add a separate `imageCaptionSlot` to avoid awkwardness with zooming on the text. but also maybe it's not that much of a problem to be able to zoom-click on the text?

I think either of these would be ok, and if you have ideas I'd be happy to hear them.

src/web/tutorial/ZoomableImage.tsx
```diff
@@ -0,0 +1,66 @@
...
+}) => {
+  const [modalOpen, setModalOpen] = useState(false);
+  return (
+    <>
+      {/* eslint-disable-next-line jsx-a11y/alt-text */}
+      <Image {...imageProps} style={{ cursor: "zoom-in" }} onClick={() => setModalOpen(true)} />
+      <Modal
+        open={modalOpen}
+        onClose={() => setModalOpen(false)}
+        sx={{ zIndex: 9999999, height: "100%" }}
```
keyserj: We've been moving away from using style/sx in favor of tailwind styling via `className` where possible, let's try to do that. Note that we do have to use sx or style in spots where we're using a variable's value as the CSS e.g. `maxWidth: imageProps.width`

```suggestion
        className="z-[9999999] h-full"
```

keyserj: Also regarding this 9999999. I'm guessing this is because we want the modal to appear in front of the tutorial, which is also in a modal? Can we add a `const tutorialZIndex = 1000` and use `style={{ zIndex: tutorialZIndex + 1 }}` just so it's explicit why we're using that number? (also note that this would mean we can no longer use `className` for this because our zIndex is set using a variable's value)

src/web/tutorial/ZoomableImage.tsx
```diff
@@ -0,0 +1,66 @@
...
+            position: "absolute",
+            top: "50%",
+            left: "50%",
+            transform: "translate(-50%, -50%)",
+            maxHeight: "90vh",
+            maxWidth: imageProps.width,
+            width: "fit-content",
+            height: "auto",
+            bgcolor: "background.paper",
+            boxShadow: 24,
```
keyserj: Is this shadow doing anything? The Material UI already provides a shadow for the background

src/web/tutorial/ZoomableImage.tsx
```diff
@@ -0,0 +1,66 @@
...
+          </Typography>
+          {/* eslint-disable-next-line jsx-a11y/alt-text */}
+          <Image {...imageProps} style={{ margin: 0 }} />
+          <Typography
+            variant="body2"
+            className="p-4 whitespace-pre-wrap"
+            sx={{ maxWidth: imageProps.width }}
+          >
+            {modalText}
+          </Typography>
```
keyserj: Let's actually just mimic the step content by putting this typography above the image. Some of the text refers to the image "below" 😅 , and I think it looks slightly better being consistent with the step content structure.

src/web/tutorial/ZoomableImage.tsx
```diff
@@ -0,0 +1,66 @@
...
+        <Box
+          sx={{
+            position: "absolute",
+            top: "50%",
+            left: "50%",
+            transform: "translate(-50%, -50%)",
+            maxHeight: "90vh",
+            maxWidth: imageProps.width,
+            width: "fit-content",
+            height: "auto",
```
keyserj: Hmm it seems like sometimes the image isn't stretching to fill up the space, particularly when the width of the screen is smaller? Though when I'm on widescreen it _does_ seem to expand. See the image below. The issue seems to be really emphasized on mobile-sized screens.

Image

keyserj: Also, when I _am_ on widescreen, it seems like the text is cut off at the bottom. See the image below. I think the sizing can be tough to get right because we want the image to expand as much as possible up to 90-ish vh/vw, but the box shouldn't have extra space around the image if the image is small. Let me know if you need ideas for this, I can try to take a look if need be.

Image

keyserj (commented):
Hey thanks for taking this on! 🙂 This is a good start.

src/web/tutorial/steps/addingNuance.tsx
henrybarl0w: I've implemented a form of the second option and just have a modal within the `StepContent` file effectively removing the need for the `ZoomableImage` component.  I've added a `imageCaption` but it can very easily not be used.

src/web/tutorial/ZoomableImage.tsx
henrybarl0w: I've done this though I'm not sure exactly what the tutorial z-index actaully is, as unless the zoom modal's z-index is around 100,000 it renders below. So I've just set `const tutorialZIndex = 100000`

src/web/tutorial/ZoomableImage.tsx
henrybarl0w: I've manged to sort of fix this though, small images still remain limit by their size and do not expand.

henrybarl0w: This is one of the hardest things to fix, for some unknown reason the class w-fit doesn't seem to work and text still expands wider than the image. I've made some changes but this issue still persists.

src/web/tutorial/ZoomableImage.tsx
keyserj: Oh yeah good catch. If you use browser devtools to look at the styles of the tutorial div, it does say that its z-index is 100000

src/web/tutorial/ZoomableImage.tsx
keyserj: Eh I think that's much better, it's good enough for me that the text isn't getting cut off anymore! Nice work

src/web/tutorial/steps/addingNuance.tsx
keyserj: I mainly had suggested the `imageCaption` as separate so that it could be _outside_ of the click-to-zoom effect. Can you see if you can get that to work? The main awkwardness is particularly when there's a caption with a URL in it - when you click the URL, the URL opens but _also_ the zoom happens, which is unexpected.

I think we probably will need another div so that we can have the sizing/centering around the image slot + image caption, and then the on-click + zoom-cursor div (with some of the sizing CSS duplicated) would be _just_ around the image slot, not the image caption.

Don't bother spending too much time on this though, I think it's not a huge deal and I could see it being a pretty annoying thing to get right.

src/web/tutorial/StepContent.tsx
```diff
@@ -30,25 +33,60 @@ export const StepContent = ({
...
+            >
+              <Close />
+            </IconButton>
+            <Typography variant="h5" className="p-4">
+              {stepTitle}
+            </Typography>
+            <Typography variant="body2" className="p-4 pt-0 whitespace-pre-wrap">
+              {text}
+            </Typography>
+            <div className="flex w-full items-center justify-center [&_img]:h-auto [&_img]:max-h-[calc(95vh-12rem)] [&_img]:w-auto [&_img]:max-w-[90vw] [&_img]:object-contain">
```
keyserj: I think this should be `flex-col` in case we pass multiple images into the slot, like this:

Image

keyserj (commented):
Apologies for taking a while to review this! This is looking good, just a few more things to spruce up.

src/web/tutorial/steps/addingNuance.tsx
henrybarl0w: I've manged to fix this, just required an extra div as you suggested.

keyserj (approved):
Great work on this, thanks for your contribution! 🚀 

Image

PR #975, opened 2026-07-22
fix: keep topic upload input mounted while reading

### LINKED ISSUE #972: Cannot upload a topic file in Firefox
### Describe the bug

Using any topic JSON, if you go to "More Actions > Topic > Upload", and select a json file, in Firefox, the topic is not successfully uploaded, and no error is thrown.

### To Reproduce

_No response_

### Expected behavior

Either the topic is uploaded, or an error is displayed.

### Additional context

Originating Discord thread https://discord.com/channels/1057707973482401892/1057707973482401898/1506851457457197086

### Technical ideas

_No response_

w3lld1:
Fixes #972

### Description of changes

- I keep the topic upload input mounted until the selected file has been read and applied.
- I close the menu only after a successful upload, avoiding Firefox invalidating the selected file while it is still being read.

### Validation

- `npm exec prettier -- --check src/web/topic/components/TopicWorkspace/MoreActionsMenu.tsx`
- `npm exec eslint -- src/web/topic/components/TopicWorkspace/MoreActionsMenu.tsx`
- `npm run check-types`
- `npm run test -- --run` (119 tests passed)
- `git diff --check`

### DIFF
--- a/src/web/common/components/Menu/FlyoutSubMenu.tsx
+++ b/src/web/common/components/Menu/FlyoutSubMenu.tsx
@@ -146,6 +146,7 @@ export const FlyoutSubMenu = forwardRef<HTMLLIElement | null, Props>(
         </MenuItem>
         <Popper
           open={open}
+          keepMounted
           anchorEl={menuItemRef.current}
           placement="right-start"
           style={{ zIndex: 1300 }}
--- a/src/web/topic/components/TopicWorkspace/MoreActionsMenu.tsx
+++ b/src/web/topic/components/TopicWorkspace/MoreActionsMenu.tsx
@@ -49,7 +49,14 @@ import {
   useMediaQuery,
   useTheme,
 } from "@mui/material";
-import { type Dispatch, type SetStateAction, useState } from "react";
+import {
+  type Dispatch,
+  Fragment,
+  type RefObject,
+  type SetStateAction,
+  useRef,
+  useState,
+} from "react";
 
 import { hasComments, resetComments } from "@/web/comment/store/commentStore";
 import { HelpIcon } from "@/web/common/components/HelpIcon";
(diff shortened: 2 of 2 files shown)


keyserj:
Hey @w3lld1 thanks for picking this up, welcome to Ameliorate 🙂

It's a good point to think about unmounting issues here, but this change doesn't seem to fix this issue, see video below. (left: Firefox with this PR's code, right: Chrome with production code)

https://github.com/user-attachments/assets/39874580-d456-43b0-aea5-18d7fca72406

What do you think? I think to debug further you could throw in some console logging on Firefox locally in `uploadFile` to see if `event.target.files` has anything.

w3lld1:
You're right—the first patch only delayed our explicit `handleClose()`, but the desktop flyout can still close and unmount its children while Firefox's file picker is open.

I pushed `c1c2bc6`, which moves the file input outside the menu/flyout lifecycle and has the Upload item trigger that persistent input. The selected file can now deliver its change event even if the menu closes while the picker is open.

Fresh validation:
- Prettier check on `MoreActionsMenu.tsx`
- ESLint on `MoreActionsMenu.tsx`
- `npm run check-types`
- full Vitest suite: 119 passed
- `git diff --check`

keyserj:
Hey @w3lld1 , great idea to move the `input` outside of the Menu Item, so that it doesn't unmount when the Menu Item is clicked and itself unmounts.

But I think this is more lines changed than needed, and I'd prefer if we can minimize the change if possible. Instead of moving the `input`, what do you think about setting `keepMounted` on the `FlyoutSubMenu`'s `Popper`? Then we'll technically be rendering all the sub menus when the more actions menu is opened, instead of just the sub menu that's displaying. But I think rendering the sub menus is pretty cheap anyway, and this is an easy way for our file `input` to remain mounted. In particular I think it's nice that this solves our issue for any other menu items that we might need similar `input` behavior for, without having to add extra behavioral wiring.

w3lld1:
Thanks — I simplified this in `50e0b1e` by keeping the `FlyoutSubMenu` Popper mounted and restoring the upload input to the Topic submenu. This keeps the fix generic while removing the extra upload wiring.

Fresh validation on the pushed commit:
- Prettier on both changed files
- ESLint on both changed files
- `npm run check-types`
- `npm run test` — 119 passed
- `git diff --check`

keyserj (approved):
Great work, thanks for contributing 🙂

PR #976, opened 2026-07-31
fix: prevent score button hover flicker

### LINKED ISSUE #561: Score button flickers when hovered
### Describe the bug

It happens quickly but if you drag the time the video instead of playing it, it's easier to see:

https://github.com/user-attachments/assets/1ef8c637-5169-4806-b0b2-9cc47e8f332a

### To Reproduce

_No response_

### Expected behavior

_No response_

### Additional context

_No response_

### Technical ideas

This happens because the score selector is portaled to the front, so a second, duplicate button is actually rendered in the portal.
- maybe we could get away without rendering the duplicate button but I'm not sure if that'd look good enough
- the first score button should probably not have a hover color change, since the selector shows up in front immediately after hover

Guflly:
Closes #561

### Description of changes

- keep the original score button's hover color unchanged while quick scoring is active
- leave normal hover behavior unchanged everywhere else

### Additional context

The portaled button still handles the selected score. This only removes the brief color mismatch before it appears.

### Validation

- `npm run check-types`
- focused ESLint and Prettier checks
- 118 of 119 Vitest tests pass locally; the remaining snapshot differs only by Windows line endings


src/web/topic/components/Score/Score.tsx
```diff
@@ -100,6 +100,7 @@ export const Score = ({ graphPartId }: ScoreProps) => {
       >
         <ScoreButton
           ref={mainButtonRef}
+          disableHoverColorChange={quickScoring}
```
keyserj, line 103: Great idea! Can you add a brief comment explaining that if we don't disable hover when quick scoring is on, that the hover darkens the button before popping up the second button in front with the old color, creating a flicker? Can link to the ticket as well

keyserj:
Bump @Guflly ! I will probably close this PR in a few days or so if it's still inactive.

